PR #90, opened 2014-11-15
Add options to EdnPacker to pass through to read-string.

ericnormand:
It should be possible to add custom tag handlers and other options for read-string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,8 @@ You're good to go! The client will automatically initiate a WebSocket or repeati
 Term          | Form                                                                  |
 ------------- | --------------------------------------------------------------------- |
 **event**     | `[<ev-id> <?ev-data>]`, e.g. `[:my-app/some-req {:data "data"}]`      |
-**event-msg** | `{:ring-req _ :event _ :?reply-fn _ :push-fn _ <...>}` (server-side)  |
+**event-msg** (server) | `{:event _ :send-fn _ :?reply-fn _ :ring-req _ <...>}`       |
+**event-msg** (client) | `{:event _ :send-fn _ <...>}`                                |
 `<ev-id>`     | A _namespaced_ keyword like `:my-app/some-req`                        |
 `<?ev-data>`  | An optional _arbitrary edn value_ like `{:data "data"}`               |
 `:ring-req`   | Ring map for Ajax request or WebSocket's initial handshake request    |
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ Here's some more unofficial/**user-submitted** examples for those interested! (*
 
 Example          | Author        | Comments                                         |
 ---------------- | ------------- | ------------------------------------------------ |
-[om-mouse]       | @tf0054       | Basic example: tracks mouse pos using Om.        |
+[om-mouse]       | [@tf0054]     | Basic example: tracks mouse pos using Om.        |
 
 #### Any other questions?
 
@@ -282,7 +282,6 @@ Otherwise reach me (Peter Taoussanis) at [taoensso.com][] or on [Twitter][]. Che
 ## License
 
 Copyright &copy; 2012-2014 Peter Taoussanis. Distributed under the [Eclipse Public License][], the same as Clojure.
-
 
 [API docs]: <http://ptaoussanis.github.io/sente/>
 [CHANGELOG_]: <https://github.com/ptaoussanis/sente/blob/master/CHANGELOG.md>
@@ -310,3 +309,4 @@ Copyright &copy; 2012-2014 Peter Taoussanis. Distributed under the [Eclipse Publ
 [jetty7-websockets-async]: <https://github.com/lynaghk/jetty7-websockets-async>
 [Socket.IO]: <http://socket.io/>
 [om-mouse]: <https://git.geekli.st/tf0054/om-mouse/tree/master>
+[@tf0054]: https://github.com/tf0054
--- a/README.md
+++ b/README.md
@@ -126,7 +126,9 @@ You'll setup something similar on the client side:
 
 ### Now what?
 
-You're good to go! The client will automatically initiate a WebSocket or repeating long-polling connection to your server.
+The client will automatically initiate a WebSocket or repeating long-polling connection to your server. Client<->server events are now ready to transmit over the `ch-chsk` channel.
+
+**Last step**: you'll want to **hook your own event handlers up to this channel**. Please see the [reference example project] for one way of doing this.
 
 #### Client-side API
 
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Closing for now. Have the change in the commit history; you're welcome to reopen if you still have interest in this going in. Cheers :-)

gphilipp:
Hi Peter & Eric, I'm using cli-time and have joda-time dates in the structures I want to pass back and forth between my Clojurescript app and my Clojure server. How can I do this either with edn or transit packer ? It seems like only this pull request would allow that (for edn at least).

ptaoussanis:
Hi Gilles,

PR welcome to reintroduce the previously-reverted commit if you can test it and confirm that it's working correctly.

Otherwise keep in mind that as a worst case you can always use `pr-str` to produce an edn string, send that over the wire, and manually `read-string` it back on the other end however you like. Does that help?

danielcompton:
@gphilipp here is some code that I'm using to add converters for Joda time dates into Transit and JSON. It's late here, I'll try and follow up with more details on it's use and whether it is another route to do what you're after.

```
(ns rethink-proxy.utils
  (:require [cheshire.generate :as gen]
            [cognitect.transit :as transit]
            [clj-time.coerce :as coerce]
            [taoensso.sente :as sente]
            [taoensso.timbre :as timbre])
  (:import [org.joda.time DateTime ReadableInstant]
           [org.joda.time.format ISODateTimeFormat]))

(defn setup-encoders
  "Add converters to turn JODA Time objects into JSON dates"
  []
  (gen/add-encoder
    DateTime
    (fn [^DateTime dt ^com.fasterxml.jackson.core.json.WriterBasedJsonGenerator generator]
      (.writeString generator (.print (ISODateTimeFormat/dateTime) ^ReadableInstant dt)))))

;; TODO: Use native Java to avoid dep on clj-time?
(def joda-time-writer
  (transit/write-handler
    (constantly "m")
    (fn [v] (-> v coerce/to-date .getTime))
    (fn [v] (-> v coerce/to-date .getTime .toString))))
```

danielcompton:
@gphilipp following up a few months later when I ran into this issue myself :)

```clj
(ns rethink-proxy.utils
  (:require [cognitect.transit :as transit]
          [taoensso.sente.packers.transit :as sente-transit])
  (:import [org.joda.time DateTime ReadableInstant]))

;; From http://increasinglyfunctional.com/2014/09/02/custom-transit-writers-clojure-joda-time/
(def joda-time-writer
  (transit/write-handler
    (constantly "m")
    (fn [v] (-> ^ReadableInstant v .getMillis))
    (fn [v] (-> ^ReadableInstant v .getMillis .toString))))

(def custom-transit-writers {DateTime joda-time-writer})

(def packer (sente-transit/->TransitPacker :json {:handlers utils/custom-transit-writers} {}))
```

danielsz:
A little contribution in this discussion, especially to the attention of boot users. :-)

I was trying to achieve exactly the same: passing custom edn tags for Joda. I'm not using transit. Not that I have anything against transit, but I'm in full control of the data passing through my system. Since my system is 100% Clojure, what flows through the various pieces (server, client) is edn. Blissful simplicity (I know transit is faster, but it's meaningless in my use case). 

I've defined custom reader tags like so: https://gist.github.com/Deraen/eb3f650c472fb1abe970#file-edn-cljx

With sente, I have not defined any flexipacker, so if my understanding is correct the default is used, which happens to be :edn. 

So far so good. 

I was debugging however an instance of 
``` clojure
clojure.lang.ExceptionInfo: No reader function for tag DateTime
    data: {:type :reader-exception}
```

The problem was that *data-readers* was not populated. I tried both with writing `data_readers.clj` and via rebiinding *data-readers*. Long story short, the issue was with boot: https://github.com/boot-clj/boot/issues/47

ptaoussanis:
Thanks Daniel, that'll be super helpful to have here as a reference for others!

danielsz:
Thank you, @ptaoussanis! 

And the good new is that the issue is fixed `system` as of `0.2.1-SNAPSHOT`. 

https://github.com/danielsz/system

danielsz:
It was a bit too soon to claim victory. 

```clojure
15-Dec-06 09:34:28 daniel-MacBookAir DEBUG [taoensso.sente] - Bad package: [[:twitter-fu/api {:update-preferences {:shop-name "test", :accounts {:twitter "bellybag1" :legacy {:end #DateTime "20151205T000000.000Z", :provider "shopify"}}, :total-items 5}}] "3cc733"] (clojure.lang.ExceptionInfo: No reader function for tag DateTime {:type :reader-exception})
```

Sente is unaware of the custom tag which has been registered here:

```clojure
(cljs.reader/register-tag-parser! "DateTime" reader-str->datetime)
```

Can you advise?

ptaoussanis:
Hi Daniel,

Please make sure that you're using `[com.taoensso/encore "2.26.1"]` or newer; some work went in there recently to try pave over issues with ClojureScript's reader tag support. The next Sente release will auto-bump this dependency but it's not included with Sente 1.7.0-RC1 yet.

If that's no help, could you possibly try confirm whether the issue still exists when not using Boot?

Thanks!

danielsz:
Hi Peter, 

Yes, that was it. No such problem with `[com.taoensso/encore "2.26.1"]`. Thank you!

Am I allowed to rant?  

Having old versions of the two auxiliary libraries `encore` and `timbre` in my classpath is something of a fixture. I remember old versions of Timbre giving me trouble (although that may have been in conjunction with Carmine). Besides being a potential source of problems, it is a difficult one to debug from a user's perspective. I never think of checking the latest versions, because these are no part of my project dependencies, and my expectation is that latest Sente is, you know, the latest. Now I see Sente ships with `[com.taoensso/encore "2.18.0"]`. Why is that? Do you expect users to track versions of the project's dependencies themselves? I mean, is there a reason for this state of affairs?

End of rant.

ptaoussanis:
> Yes, that was it. No such problem with [com.taoensso/encore "2.26.1"]. Thank you!

No problem. Will try cut a new Sente release with the fix in the next few days, just have my hands a bit full atm.

> Having old versions of the two auxiliary libraries encore and timbre in my classpath is something of a fixture.

Fixture?

>  I never think of checking the latest versions, because these are no part of my project dependencies, and my expectation is that the latest Sente is really the latest (including dependencies).

You don't need to think of checking the latest versions (that'd be documented otherwise), your expectation is correct that libs should pull in the necessary dependencies.

In this case: you brought a Sente issue to my attention for the first time. I suspected that maybe a new version of Encore would help. You just confirmed that. So my intention is to bump the Encore dependency for Sente's next overdue release, fixing the issue. This way you get the fix sooner than if you'd have had to wait for the next Sente release so not sure what state of affairs here warrants ranting?

Does that make sense / seem reasonable? :-)

danielsz:
Yes, absolutely. I'm sorry for the tone of the rant. 
If I had any trouble with your otherwise excellent libraries, it was with the auxiliary ones (timbre and encore). It is my responsibility to report it when it happens, not as a result of past frustrations.

ptaoussanis:
No problem, happy you reported this :-)

PR #92, opened 2014-12-19
Update example for 1.3.0rc1

pandeiro:
Hi Peter,

In my quest to better understand how `sente` works, I've been running through examples. I noticed that the project's own example was using some outdated dependencies and configuration. I've tested these changes (to the example-project only) and it works for me here.


ptaoussanis:
Closing, fixed on master.

PR #109, opened 2015-02-25
exclude cljx files from jar

sritchie:
Prevents cljx files from making it into the published jar. This may fix an issue I'm having with execution at the emacs repl. These shouldn't get passed along, in any case.

### DIFF
--- a/example-project/src/example/my_app.cljx
+++ b/example-project/src/example/my_app.cljx
@@ -74,7 +74,7 @@
      :stop-fn (fn [] (http-kit-stop-fn :timeout 100))}))
 
 ;;; Immutant
-;; #+clj (def web-server--adapter taoensso.sente.server-adapters.immutant/immutant-adapter)
+;; #+clj (def web-server-adapter taoensso.sente.server-adapters.immutant/immutant-adapter)
 ;; #+clj
 ;; (defn start-web-server!* [ring-handler port]
 ;;   (println "Starting Immutant...")
--- a/project.clj
+++ b/project.clj
@@ -7,6 +7,7 @@
             :distribution :repo
             :comments "Same as Clojure"}
   :min-lein-version "2.3.3"
+  :jar-exclusions [#"\.cljx|\.DS_Store"]
   :global-vars {*warn-on-reflection* true
                 *assert* true}
 


ptaoussanis:
Hey Sam.

Hmm, I'm not sure about this. I tend to like the .cljx files around while browsing source, etc. - but maybe I'm unusual?

In any case I don't believe this should be causing _errors_. Even if you have .cljx files around, nothing but the Cljx processor should be trying to compile them in a properly configured environment.

Have you made any changes to your environment recently? If you're just seeing errors from the REPL, I'd start by checking that your Cljx REPL middleware is configured correctly.

sritchie:
Hey, saw your other note. I'm going to do a full clean-fest tonight and see if I can't make this all work.

sritchie:
This error makes no sense, but with a full clean I'm still getting it. I've upgraded Sente to 0.4.0-SNAPSHOT, and that's it. Let me look at a dependency diff to see what's up.

sritchie:
This is with tools.repl 0.2.7 and cljx 0.6.0.

sritchie:
Works great if I exclude these cljx files. I'm going to publish a version now to at least get unblocked.

ptaoussanis:
Quick heads-up that `[com.taoensso/sente "1.4.0-beta1"]` is currently the latest version (newer than the SNAPSHOT) - https://github.com/ptaoussanis/sente/releases

sritchie:
Yup, all my testing today was with that beta version.

sritchie:
I also confirmed that I was getting this issue at the "lein repl" in addition to my cider repl.

ptaoussanis:
So, sorry, just remind me what error you're getting exactly and where?

What's generating the error? How is a Cljx pre-processor tag getting into your REPL?

sritchie:
No idea how a preprocessor tag is getting into my repl.

When I inspect the classpath and look at all files in all jars, Sente's are the only CLJX copies of clojure and clojurescript files that I'm actually including.

This is the error I get when I run "lein repl" and call `(require 'racehub.server)` (which transitively requires Sente):

```clojure
CompilerException java.lang.RuntimeException: No reader function for tag +clj, compiling:(taoensso/sente:606:89) 
```

The error went away when I published a local version of Sente with cljx files excluded. I could also kill it by C-c C-k'ing the main sente.cljx file manually.

ptaoussanis:
Okay, so - you're running Emacs, Cider, and Cljx 0.6.0?

Have you got the [Cljx REPL middleware installed](https://github.com/lynaghk/cljx#repl-integration)?

What happens if you open up a new dummy .cljx file and evaluate something like this ns?:
```clojure
(ns my-dummy-cljx-ns)
#+clj (println "hello world")
```

If your Cljx middleware is working correctly, you should see the text go to your stdout (no error).

sritchie:
Yeah, most of my project is cljx, and has been since the beginning. I've got the middleware installed, and everything works great at the repl. Also, remember that this shows up outside of emacs and cider. I can recreate with "lein repl", just by calling require.

ptaoussanis:
> Yeah, most of my project is cljx, and has been since the beginning. I've got the middleware installed, and everything works great at the repl.

Sorry if you've mentioned this already, busy juggling 10 projects atm. Just to clarify:
1. Everything works fine at the Emacs REPL.
2. `lein repl` followed by a `require` call triggers the error.
3. Does anything else trigger the error? What about `lein check`?
4. Do you see the same error when running the reference example project in the same way?

sritchie:
1. No - Emacs repl fails as well when I run C-c C-k. I noted that "lein repl" gives me the same error to show that it's not Cider or Emacs at fault.
2. yup.
3, 4. trying now

ptaoussanis:
> No - Emacs repl fails as well when I run C-c C-k.

So even with a dummy ns file **with a .cljx extension** like this:
```clojure
(ns my-dummy-cljx-ns)
#+clj (println "hello world")
```

C-c C-k in Emacs results in an error? That shouldn't be. That seems to be pointing to a misconfiguration somewhere - either with the middleware or maybe the output file config.

ptaoussanis:
About to head out to dinner - will keep an eye on this thread from my phone. Keep the debug info coming if you can, and I'll try get back to you in a couple hours when I'm back at my hotel.

Good luck, cheers! :-)

sritchie:
Haha, no, sorry. Let me clarify - everything about my CLJX setup works beautifully. I can C-c C-k CLJX files, including sente. I can compile my own stuff. I can enter that dummy ns file just great at the repl. So good.

I get the error I pasted above when I bump my sente version up to 1.4.0-beta1 and try to C-c C-k (or just straight-up require) a normal clj file that requires sente somewhere.

Also, as I noted, Sente's are the only cljx files on my classpath.

sritchie:
If I go ahead and publish my own version of Sente that excludes those CLJX files from the classpath (like all of my other project dependencies that have CLJX files, like Prismatic's Schema, seem to have done) then everything works beautifully too.

ptaoussanis:
Okay, so let's summarise for a moment:

**With Sente v1.3.0**:
* Emacs C-c C-k in any .cljx file works w/o errors.
* Emacs C-c C-k in any .clj file using Sente works w/o errors.
* `lein repl` + requiring any .clj file using Sente works w/o errors.

**With Sente v1.4.0-beta1**:
* Emacs C-c C-k in any .cljx file works w/o errors.
* Emacs C-c C-k in any .clj file using Sente generates an error:
```clojure
CompilerException java.lang.RuntimeException: No reader function for tag +clj, compiling:(taoensso/sente:606:89)
```
* `lein repl` + requiring any .clj file using Sente generates the same error.

**Further observations**:
* This happens even after a `lein clean`.
* It is reliably, reproducibly the case that Sente v1.4.0-beta1 generates the error and v1.3.0 does not.
* Adding a .cljx entry to Sente's project.clj `:jar-exclusions` prevents these errors.

----

**My feedback, assuming I have all of that correct:**

Yes, adding a .cljx entry to Sente's project.clj `:jar-exclusions` may prevent these errors. And other libs may be choosing to exclude .cljx files from their jars as a matter of course. My current feelings on this:

1. I think I'd prefer to keep .cljx files in the jar if possible since I tend to like them there and their inclusion shouldn't[1] be harmful besides increasing the jar file size slightly.
2. I'd prefer to have an explanation for why you weren't seeing errors with Sente v1.3.0 but are with Sente v1.4.0-beta1. Side-stepping the .cljx files to avoid the error seems like a treatment of symptoms without understanding the cause. Why would anything be trying to evaluate a pre-processed .cljx file? That doesn't seem right.

[1] I'm open to being corrected here, of course.

Does this make sense?

**Some follow-up debugging questions for you**:

1. Do you see any errors when running the reference example project in the same way against v1.4.0-beta1?
2. Did anything else change in your tool chain, project.clj, or build procedure when you migrated from v1.3.0 to v1.4.0-beta1?

sritchie:
Hey, thanks for the summary. I don't get the same error when I run `lein repl :headless` and then `C-c C-k` inside of the example project's main file. I'll see if I can try moving some other dependencies from my project over to the example project to try and activate the issue. Strange stuff.

ptaoussanis:
> I'll see if I can try moving some other dependencies from my project over to the example project to try and activate the issue. Strange stuff.

I'd start by looking at any unusual build tools, etc. Nothing should be touching a .cljx file besides the Cljx processor. Seems like there's only two possible causes for the error you're seeing:
1. The Cljx processor is outputting files with #+clj / #+cljs tags still present => a bug or misconfiguration.
2. Something is trying to evaluate .cljx files which shouldn't be.

Note that the error you're seeing apparently doesn't mention any specific file extension:
`CompilerException java.lang.RuntimeException: No reader function for tag +clj, compiling:(taoensso/sente:606:89)`

It'd be helpful to try pinpoint:
A. What file this tag is present in.
B. What process is currently evaluating the file when the tag is encountered.

To debug, I'd start by trying to rule out cause 1 as a possibility.

Grep through all the .clj and .cljs files looking for a #+clj / #+cljs tag. If you find any .clj or .cljs files with these tags, you know something's off with the Cljx processing.

If you _don't_ find any .clj or .cljs files with these tags (i.e. tags are only present in pre-processed .cljx files as they should be) - then it'll be a case of trying to figure out what piece of software in your tool chain is attempting to evaluate a pre-processed .cljx file.

Does that help / make sense?

sritchie:
Yup, the debugging path totally makes sense. I've gone through all of that, and the tag's present at that line in sente.cljx, which as I noted is one of the only cljx files on my classpath that's being evaluated. #2 is of course what's going on, I'm just not sure what in my REPL stack is trying to do this.

sritchie:
Okay, I've tried

- Moving all my dependencies over to the example project
- Attaching the piggieback middleware to the example project
- upgrading cljs versions over there
- Moving my code over to the example project

And I can't recreate this in the example project. I'm going to move on from debugging this for now, use my published version with no cljx files and hope that inspiration hits later. Happy to help debug if someone else reports the problem.

ptaoussanis:
> I'm going to move on from debugging this for now, use my published version with no cljx files and hope that inspiration hits later.

Understood. Will let you know if I think of anything specific you could try or if I encounter this behaviour anywhere else in future.

Thanks for working through debug with me.

PR #113, opened 2015-03-11
Updated example in README.md

jgrimes:
Updated example in README.md to use the updated make-channel-socket! which takes an adapter


ptaoussanis:
Hi Justin, thanks for catching this! Much appreciated :-)

PR #119, opened 2015-04-05
Added code snippet with `wrap-keyword-params` and `wrap-params`

bensu:
Since it is a requirement to have `wrap-params` and `wrap-keyword-params` for sente to work I thought it would be better if it was documented in the README.

Hope this helps


ptaoussanis:
Hi Sebastian,

Thanks for this - sorry for the delay getting back to you!

bensu:
No need to apologize! Thank you for Sente. It works great.

PR #122, opened 2015-04-29
README fix: wherefore means why, not where.

ane:
http://www.oed.com/view/Entry/228218


ane:
In case the link doesn't work: http://en.wiktionary.org/wiki/wherefore

ptaoussanis:
That's my sense of humour apparently not communicating too well ;-)

ane:
Oh. Ahem.

ptaoussanis:
No problem, never hurts to check :-)

PR #128, opened 2015-06-05
Bootified example

danielsz:
Hello Peter, 

It took me a while, but here I am with a Bootified example app. 

The example app remains the same, the toolchain is different. 

If you're OK with the principle of having a Leiningen and a Boot example app side by side, I'll add READMEs on both to explain usage.

Lemme know :-)


ptaoussanis:
Hey Daniel, thanks a lot for this!

Not familiar with Boot at all - is it really necessary to duplicate the entire example's source code? Would ideally prefer if we could share the same application code (which I update fairly regularly).

Are there actually changes in the application code to support Boot?

danielsz:
superseded by #129

Issue #50, opened 2014-07-01
Why does sente restrict ws communication to same-domain?


nodename:
I was hoping to use sente to provide web-socket access to a server from arbitrary external domains, not obviously an unreasonable use case (chat server for example). Could there be a workaround?

danielcompton:
I'm not 100% clear on how to use this extension. Do we have to implement our own `chsk-url-fn`, or can we pass a map of the things we want somewhere?

ptaoussanis:
Hi Daniel,

You can use the `:chsk-url-fn` option to override the default default chsk url fn:
```clojure
(def default-chsk-url-fn
  "(ƒ [path window-location websocket?]) -> server-side chsk route URL string.

    * path       - As provided to client-side `make-channel-socket!` fn
                   (usu. \"/chsk\").
    * websocket? - True for WebSocket connections, false for Ajax (long-polling)
                   connections.
    * window-location - Map with keys:
      :href     ; \"http://www.example.org:80/foo/bar?q=baz#bang\"
      :protocol ; \"http:\" ; Note the :
      :hostname ; \"example.org\"
      :host     ; \"example.org:80\"
      :pathname ; \"/foo/bar\"
      :search   ; \"?q=baz\"
      :hash     ; \"#bang\"

  Note that the *same* URL is used for: WebSockets, POSTs, GETs. Server-side
  routes should be configured accordingly."
  (fn [path {:as window-location :keys [protocol host pathname]} websocket?]
    (str (if-not websocket? protocol (if (= protocol "https:") "wss:" "ws:"))
         "//" host (or path pathname))))
```

I.e. you can replace that function with your own function of the same arguments that produces URLs of the form you prefer. Does that make sense?

luposlip:
@danielcompton I'm using it with Sente 1.5.0-beta2 (https://clojars.org/com.taoensso/sente).

My code looks like this:
```
(sente/make-channel-socket! "/chsk"
   {:type :auto :packer packer
    :chsk-url-fn (constantly "ws://host:port/chsk")})
```
In this case the first parameter - "/chsk" in the above example - doesn't make much sense though, since it seems to be ignored completely and replaced by the :chsk-url-fn.

@ptaoussanis perhaps it would make sense to have  a :chsk-host-fn instead, and then concatenate the first url param and the host (if present)?

ptaoussanis:
Hi Henrik, thanks for the example.

[...]

PR #136, opened 2015-06-14
Add host-chsk-url-fn

danielcompton:
It would be helpful for people using cross domain channel sockets to have a simplified `chsk-url-fn` for making connections to different hosts. This PR adds the function, and documentation on how to use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 > This project uses [Break Versioning](https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md) as of **Aug 16, 2014**.
 
+## v1.5.0 - 2015 Jun 11
+
+> This is a non-breaking maintenance release
+
+* **New**: support Ajax CORS via new `:with-credentials?` opt [#130 @bplatz]
+* **Fix**: bad missing-middleware error logging call format
+* **Implementation**: update dependencies
+
+
+```clojure
+[com.taoensso/sente "1.5.0"]
+```
+
+
 ## v1.4.1 - 2015 Mar 11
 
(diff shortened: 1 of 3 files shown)


danielcompton:
If you'd prefer fewer functions in the main Sente ns, the other option I thought of was making `default-chsk-fn` take a map of config and return a function. The function would override any config provided by `window-location`, with the initial config provided. This would be a breaking change though, and more complex to use.

ptaoussanis:
Hi Daniel, thanks for this.

As I mentioned last time this came up (in #50), I'm not convinced it makes sense to add any specialised code for the use case you're describing.

An appropriate chsk-url-fn can usually be expressed more flexibly and transparently as a small inline fn, no?

What problem are you actually trying to solve here? Is it that you feel users won't understand how to write an appropriate chsk-url-fn? Could that not better be solved by adding a short cross-domain example to the `default-chsk-url-fn` docstring? (Just trying to understand your motivation better).

Cheers :-)

danielcompton:
I thought that users may not understand how to write the chsk-url-fn correctly. A docstring would be fine too if you'd prefer that.

ptaoussanis:
Okay, going to simplify this. Will keep `:chsk-url-fn` for the moment as an advanced/undocumented feature.

For the simpler common-case that you're describing, have added a `:host` key to the client-side `make-channel-socket!` options. Does that handle your case okay?

danielcompton:
Sure, if you prefer that, I'm happy with any of the options.

ptaoussanis:
Great, closing this then. Change is up on the dev branch and `[com.taoensso/sente "1.6.0-alpha2"]` on Clojars. Cheers :-)

PR #140, opened 2015-06-14
Tidy up conditional logic

danielcompton:
Reading the Sente source I found a few places weren't clear, this PR reorders or reformats the conditional logic, while keeping the behaviour the same.

* Reorder conditional expressions to make logic more clear
* Reindent lines
* Split if's over 3 logical lines


ptaoussanis:
Hi Daniel, I do prefer the current formatting to be honest - thank you though! :-)

PR #142, opened 2015-06-17
Expose uid on ev-msg's

danielcompton:
It would be handy to be able to get the `uid` directly on the `ev-msg`. There are a few caveats though:

* I'm not sure how to get this for AJAX POST's (marked as a TODO in the commit)
* I'm not sure of the security implications of this. I think it's solid, because it's using the uid returned by `user-id-fn` on the server but I'd like some other minds on this too.

This isn't ready to merge yet, open for discussion :).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 > This project uses [Break Versioning](https://github.com/ptaoussanis/encore/blob/master/BREAK-VERSIONING.md) as of **Aug 16, 2014**.
 
+## v1.5.0 - 2015 Jun 11
+
+> This is a non-breaking maintenance release
+
+* **New**: support Ajax CORS via new `:with-credentials?` opt [#130 @bplatz]
+* **Fix**: bad missing-middleware error logging call format
+* **Implementation**: update dependencies
+
+
+```clojure
+[com.taoensso/sente "1.5.0"]
+```
+
+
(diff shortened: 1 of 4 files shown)


ptaoussanis:
Hey Daniel, this would be a nice addition - thanks!

Will comment on the PR in a sec.

danielcompton:
I've updated the PR with the logic you suggested. It won't work because `client-id` isn't available here, but I'm guessing you'll fix that when you refactor this.

danielcompton:
Ah, got client-id sorted.

ptaoussanis:
Thanks looks ready to merge if you can rebase against the dev branch, squash the commits, and skip the change to project.clj?

Much appreciated!

danielcompton:
I'm a little confused with the state of the git branches, dev and master seem to have diverged (perhaps because of a rebase?). Theres https://github.com/ptaoussanis/sente/commit/c36bd11596cea4e325f4889b636ffa3e983e0dbc and https://github.com/ptaoussanis/sente/commit/fbe61764fd79079f37299f80d638a52f13eedd7d which both have the same contents but different SHA's. Do you want to make those two match before I go further? I think that might be why the it looks like I'm trying to add your changes in my pull request too.

danielcompton:
Possibly I've made a new pull request #147 which merges these changes into dev, not master.

PR #144, opened 2015-06-20
Add a run command to the example project

estsauver:
Adds a command to build the project and run it. Simplifies setting
up a working version for people who are not running compatible cider
setups, or who don't want to run it from there repl.


danielcompton:
Why did you add the :gen-class?

estsauver:
You're right, it doesn't need the gen-class. I'm also removing the project.clj alias. It only *really* needs the main bit.

ptaoussanis:
Thanks, just tweaked this so it's `#+clj` only!

PR #145, opened 2015-06-20
Let TransitPacker accept transit r/w opts

estsauver:
TransitPacker should accept an options argument for the reader/writer.
This allows the set of messages that tranist supports to be extended
via adding additional handlers.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject com.taoensso/sente "1.5.0"
+(defproject com.taoensso/sente "1.6.0"
   :author "Peter Taoussanis <https://www.taoensso.com>"
   :description "Clojure channel sockets library"
   :url "https://github.com/ptaoussanis/sente"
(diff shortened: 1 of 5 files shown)


project.clj
```diff
@@ -1,4 +1,4 @@
-(defproject com.taoensso/sente "1.5.0"
+(defproject com.taoensso/sente "1.6.0"
```
estsauver: Not sure if this is correct, but it seemed like Break Versioning would imply a bump.

ptaoussanis:
Hi Earl, thanks for this!

In principle keen to support the transit opts. Few things we'd need to clean up with this PR though - will get back to you when I can.

> Not sure if this is correct, but it seemed like Break Versioning would imply a bump.

Thanks- you can ignore things like the version number and changelog, etc. - I'll update those when cutting a new release.

ptaoussanis:
Closing, have added support on the dev branch. Thanks again!

PR #156, opened 2015-08-27
Remove extra ampersand from cljs make-channel-socket bindings

danielcompton:
I think it's a fluke that destructuring with two &'s works in ClojureScript, it's not legal in Clojure. This patch removes the extra ampersand, and in the process fixes a bunch of resolution issues in Cursive Clojure.


ptaoussanis:
Awesome, thank you Daniel! Will try cut a hotfix release shortly.

ptaoussanis:
On Clojars as `[com.taoensso/sente "1.7.0-alpha3"]`

PR #158, opened 2015-08-28
Add custom query params to make-channel-socket!

### LINKED ISSUE #135: Passing an authentication token in the websocket upgrade request
I'm trying to use Sente where the Sente server is on one domain, the webapp files are served from another one, and a third server is providing an authentication token. I've got things mostly working, but I'd like to be able to authenticate the user and assign a `uid` from the token when they connect to the channel socket. The Sente server will authenticate the user with a JWT token, not session storage.

It's not clear to me how to pass the token in the initial connection as there doesn't seem to be an extension point for me to pass any more data here: https://github.com/ptaoussanis/sente/blob/v1.4.1/src/taoensso/sente.cljx#L843-L844

Alternatively, am I meant to connect first, then authenticate second? If so, once a client is authenticated, how would I check that every subsequent channel socket message was authenticated and update the `uid` of the connection? It would be much cleaner for this to be part of the ring request, rather than kept in an external store that I have to keep in sync with Sente's connection state. 

I've seen https://github.com/seancorfield/om-sente/blob/master/src/clj/om_sente/server.clj#L112-L119, but don't particularly like having an external auth store, and it also uses Ring session storage to get the `uid`.

It's possible that I'm misunderstanding this entirely, how else would you recommend handling cross domain auth/Sente connections?

danielcompton:
This patch allows the user to pass additional query parameters in the additional web socket connection. These can be used for authentication.  Websockets don't have an API to pass any headers, so any out of band information has to go over the query params.

Because `:chsk-url-fn` has been deprecated, I think we need to have another parameter for this. A slightly different approach would be to adjust the internal API so client-id isn't passed separately to the `map->*` functions, but instead we just pass a map of params.

Fixes #135

### DIFF
--- a/src/taoensso/sente.cljx
+++ b/src/taoensso/sente.cljx
@@ -754,7 +754,7 @@
 
 #+cljs ;; Handles reconnects, keep-alives, callbacks:
 (defrecord ChWebSocket
-    [client-id url chs socket_ kalive-ms kalive-timer_ kalive-due?_ nattempt_
+    [client-id url params chs socket_ kalive-ms kalive-timer_ kalive-due?_ nattempt_
      cbs-waiting_ ; {<cb-uuid> <fn> ...}
      state_       ; {:type _ :open? _ :uid _ :csrf-token _ :destroyed? _}
      packer       ; IPacker
@@ -815,7 +815,8 @@
              (if-let [socket
                       (try
                         (WebSocket. (enc/merge-url-with-query-string url
-                                      {:client-id client-id}))
+                                      ; We want params first so people don't clobber our query params accidentally
+                                      (merge params {:client-id client-id})))
                         (catch js/Error e
                           (errorf e "WebSocket js/Error")
                           nil))]
@@ -869,7 +870,7 @@
 
 #+cljs
 (defrecord ChAjaxSocket
-    [client-id url chs timeout-ms ajax-opts curr-xhr_ state_ packer
+    [client-id url params chs timeout-ms ajax-opts curr-xhr_ state_ packer
      backoff-ms-fn]
   IChSocket
   (chsk-send!* [chsk ev {:as opts ?timeout-ms :timeout-ms ?cb :cb :keys [flush?]}]
@@ -937,6 +938,7 @@
                   :resp-type :text ; Prefer to do our own pstr reading
                   :params
                   (merge
+                    params ; We want params first so people don't clobber our query params accidentally
                     {:_          (enc/now-udt) ; Force uncached resp
                      :client-id  client-id}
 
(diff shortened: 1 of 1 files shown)


src/taoensso/sente.cljx
```diff
@@ -754,7 +754,7 @@
 
 #+cljs ;; Handles reconnects, keep-alives, callbacks:
 (defrecord ChWebSocket
-    [client-id url chs socket_ kalive-ms kalive-timer_ kalive-due?_ nattempt_
+    [client-id url params chs socket_ kalive-ms kalive-timer_ kalive-due?_ nattempt_
```
danielcompton: `client-id` and `params` could be combined.

ptaoussanis:
Hi Daniel,

Thanks for the clear PR, much appreciated! :+1: :-)

Am I correct in understanding that this change is just to make it easier to include query-params in the chsk URL? You're aware that the URL/path can already contain query params, right?

ptaoussanis:
This looks good to me btw, will merge + cut a new release if you give me the go-ahead.

> client-id and params could be combined.

Think probably better to keep them separated as you have. They're semantically different (params are user-space, client-id required for the implementation).

danielcompton:
> Am I correct in understanding that this change is just to make it easier to include query-params in the chsk URL? You're aware that the URL/path can already contain query params, right?

Rereading https://github.com/ptaoussanis/sente/issues/135#issuecomment-118776050 earlier today I clicked on what you were meaning. However now that chsk-url-fn has been deprecated I don't see any other way to influence the URL that is used for the connection?

>  if you give me the go-ahead.

:shipit: 

> They're semantically different

That's how I felt too.

Thanks!

ptaoussanis:
> However now that chsk-url-fn has been deprecated I don't see any other way to influence the URL that is used for the connection?

Oh, just with the `path` argument. So if your path is usually `/chsk`, you could change it to something like "/chsk/?email=foo%40example.com", etc.

Does that make sense?

Sorry if I wasn't too clear before. In any case, happy with this PR since it definitely makes things easier.

> :shipit:

Is that... a gopher? :-)

ptaoussanis:
Oh btw see you adjusted this to target the dev branch, thanks!

danielcompton:
That's the shipit emoji :)

ptaoussanis:
> That's the shipit emoji :)

Perhaps telling that I've never seen it before ;-)

Okay, `[com.taoensso/sente "1.7.0-alpha4"]` is up on Clojars. Does this fully resolve #135 for you? Okay to close that?

danielcompton:
Yep, closed it now. Thanks!

ptaoussanis:
Quick update: just pushed `[com.taoensso/sente "1.7.0-alpha5"]` since we missed a case: the params weren't being included in Ajax-based callback requests. Should be okay now.

danielcompton:
Thanks, and sorry about that.

ptaoussanis:
Oh, no problem - easy to miss. PR was very helpful, wouldn't have gotten to this otherwise.

Issue #150, opened 2015-07-24
Connection Leak


bdollard:
I've found a connection leak when using sente with Immutant in AJAX mode. I'm not sure if it is a sente bug or an immutant bug.

Step to reproduce.

1. Clone https://github.com/bdollard/sente  (this has just been modified to use immutant and ajax and disable broadcasting) and run the example project 
2. leave the browser window open and and observe the number of connections in CLOSE_WAIT status gradually grow as longpolling connections time out.  

Eventually the open files limit can be reached, making the server non-responsive.  I did a little debugging, and it looks like  [on-close handler] (https://github.com/ptaoussanis/sente/blob/master/src/taoensso/sente/server_adapters/immutant.clj#L22-L24) is not getting called for connections that time out. 
This may be the cause of #146

tobias:
@bdollard  - I see your example uses Immutant 2.0.1. Have you tried 2.0.2? It has a [fix](https://issues.jboss.org/browse/IMMUTANT-564) that specifically addresses the on-close handler not getting called, and may fix the issue.

bdollard:
I just updated the fork to use 2.0.2, I does not appear to make any difference.

tobias:
I can recreate the issue with your example, even with 2.0.2. This is most likely a bug in Immutant itself, not Sente. I created https://issues.jboss.org/browse/IMMUTANT-572 to track it there, and can take a look on Monday. If the fix is fairly straightforward, I should be able to release 2.0.3 pretty quickly.

ptaoussanis:
Thanks for grabbing this @tobias! Please feel free to ping me if it turns out the trouble might be on Sente's end.

Cheers :-)

tobias:
It turns out this is an issue with a collision of assumptions made by
Undertow (the web server inside Immutant) and Sente.

Sente assumes that if the client closes a connection, the server will
detect and close as well, without any interaction from Sente
itself. Http-kit does this, I believe because it calls NIO's select()
in a loop, which may notice and close the CLOSE_WAIT sockets
immediately. When Http-kit closes the socket, that bubbles up and
triggers Sente's :on-close handler to properly remove the channel from
the connection set.

Undertow, however, treats connections differently - if you mark the
request as async, it doesn't check it again (by default) until you try
to read from, write to, or close it. At that point it will notice the
half-closed state, and close the socket fully. It assumes the caller
will eventually write to or close the request. If you never write to
the connection, then the half-closed socket is never detected (so gets
[...]

PR #159, opened 2015-08-28
Use timeouts for lp connections to prevent socket leaks [#150]

tobias:
This sets a timeout on the server-side for long-polling connections
so they get cleaned up when the client closes as part of its standard
ping mechanism or goes away.

This requires a recent Immutant incremental
build (http://immutant.org/builds/2x/) or Immutant 2.1.0, once it is
released.

### DIFF
--- a/src/taoensso/sente.cljx
+++ b/src/taoensso/sente.cljx
@@ -1005,10 +1005,16 @@
                     ; application-level auth, etc.)
     :ws-kalive-ms   ; Ping to keep a WebSocket conn alive if no activity w/in given
                     ; number of milliseconds
-    :lp-timeout-ms  ; Ping to keep a long-polling (Ajax) conn alive ''
+    :lp-timeout-ms  ; Ping to keep a long-polling (Ajax) conn alive '' [1]
     :packer         ; :edn (default), or an IPacker implementation (experimental)
     :ajax-opts      ; Base opts map provided to `taoensso.encore/ajax-lite`
-    :wrap-recv-evs? ; Should events from server be wrapped in [:chsk/recv _]?"
+    :wrap-recv-evs? ; Should events from server be wrapped in [:chsk/recv _]?
+
+  [1] If you are using Immutant and override the default :lp-timeout-ms, you will
+      need provide the same value to
+      `taoensso.sente.server-adapters.immutant/make-immutant-adapter` and use the
+      result of that function as the web-server-adapter to your server-side
+      `make-channel-socket!`."
   [path &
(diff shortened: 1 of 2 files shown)


ptaoussanis:
Just finishing up some urgent work tonight; okay if I try take a look at this Sat/Sun?

tobias:
Sure, no rush. I'm going to publish a fork on clojars in the meantime to get feedback/testing from users.

ptaoussanis:
Awesome, thanks!

ptaoussanis:
@tobias Hey Toby, you ready for me to merge this today?

How's the solution looking? Should this be considered a total solution of #150 or is there still something else that'd you'd ideally like to see Sente doing on its end to assist?

Cheers :-)

tobias:
@ptaoussanis - Peter, we're still working on getting Immutant 2.1.0
out - it will now be tomorrow at the earliest I think. I'd like to
wait for that before merging if that's okay.

tobias:
Also, this is a total solution for #150, and I'm not sure what Sente
could do to assist, other than having a baked-in solution that would
work for any server that would replace this patch.

One option would be to have a timeout-based go block on the
server-side (similar to the existing `flush-buffer!` go block) that
closes every lp connection. It would result in extraneous closes in
cases where data was sent, but would ensure all of the connections get
closed, regardless of the underlying server.

I'm happy to explore that if you are open to it - I don't know sente
well enough to know what the performance or other impacts of such a
change would be, so I would need guidance there.

If that's not a change you are interested in, I'm fine with going with
this patch, since only Immutant leaks sockets currently. If someone
provides an implementation for another server in the future, and it
also suffers from socket leaks in the same way, I'm happy to help with
a more general solution then.

ptaoussanis:
> Peter, we're still working on getting Immutant 2.1.0 out - it will now be tomorrow at the earliest I think. I'd like to wait for that before merging if that's okay.

Sure, no problem (zero rush on my end) - how about you merge whenever you're ready? Just ping me if you'd like a new release cut.

> One option would be to have a timeout-based go block on the [...]

This (or something like it) should be doable. My own inclination atm would be to wait and see first if this issue actually turns up again with any other servers (Jetty, etc.). Haven't had any time to investigate if this is something unique to Undertow, or just a pattern that some servers prefer to follow.

> I'm happy to explore that if you are open to it 

That's a kind offer, thank you. First step if you (or someone else) felt like doing any extra work on this would be digging to see if there's actually a need for any other servers.

If there is, then a Sente-side solution would likely be the way to go.

>  If someone provides an implementation for another server in the future, and it also suffers from socket leaks in the same way, I'm happy to help with a more general solution then.

Sure, thank you. Let's see if anything else turns up and proceed from there?

Issue #149, opened 2015-07-06
Switching to reader conditionals and 1.7


danielcompton:
What are your thoughts/plans for updating Sente to 1.7 and using cljc? I'm not sure how far back/how long you want to support older versions of Clojure for?

ptaoussanis:
Hey Daniel,

I like to try keep back-compatibility for at least 1 Clojure version whenever possible unless there's a compelling reason to break. Did you have something specific in mind?

Current plan is to move to .cljc sometime after Clojure 1.8, possibly. Likeliest reason to abandon that plan would be an important breaking change in ClojureScript which needs Clojure 1.7+.

Does that help / answer your question?

danielcompton:
Yeah, definitely, thanks!

danielcompton:
Now that 1.8 has been out for a while, would you be interested in updating to use reader conditionals?

ptaoussanis:
W/in the next few months potentially, sure. May I ask: is there any reason for urgency? Now that the ref. example is split (no longer uses .cljx), this seems like a strictly implementation detail, no?

Not against it if there's a reason; just want to prioritise efforts. A conversion will realistically be at least an hour or two of someone's time.

danielcompton:
I'm happy to do the work. I find cljx a bit hard to work with in Cursive, and symbol resolution in our app doesn't work well with CLJX files. There may be a few PRs coming up, and this would make it easier for me.

ptaoussanis:
Okay, I'm on board if we target Sente v1.10, i.e. after v1.9 final is out? ETA: maybe ~2 weeks?

ptaoussanis:
Apologies for the delay; haven't had any time to work on this during my trip. Will be back on Monday + should have a new release cut shortly after.

ptaoussanis:
Resolved by #242

PR #166, opened 2015-09-23
Added support for Node

whamtet:
I added a server adapter for Node.js.  In addition to the adapter I added sente-node.cljs.  This is sente.clj adapted to clojurescript on the server.  Ideally this can be merged back into sente.cljx however it will require the renaming of some functions because it breaks the implicit assumption that all cljs is on the client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,15 +154,15 @@ The client will automatically initiate a WebSocket or repeating long-polling con
 
 ===============
 
-Term          | Form                                                                  |
-------------- | --------------------------------------------------------------------- |
-**event**     | `[<ev-id> <?ev-data>]`, e.g. `[:my-app/some-req {:data "data"}]`      |
-**event-msg** (server) | `{:event _ :send-fn _ :?reply-fn _ :ring-req _ <...>}`       |
-**event-msg** (client) | `{:event _ :send-fn _ <...>}`                                |
-`<ev-id>`     | A _namespaced_ keyword like `:my-app/some-req`                        |
-`<?ev-data>`  | An optional _arbitrary edn value_ like `{:data "data"}`               |
-`:ring-req`   | Ring map for Ajax request or WebSocket's initial handshake request    |
-`:?reply-fn`  | Present only when client requested a reply.                           |
+Term          | Form                                                                   |
+------------- | ---------------------------------------------------------------------- |
+event         | `[<ev-id> <?ev-data>]`, e.g. `[:my-app/some-req {:data "data"}]`       |
+server event-msg | `{:keys [event id ?data send-fn ?reply-fn uid ring-req client-id]}` |
+client event-msg | `{:keys [event id ?data send-fn]}`                                  |
+`<ev-id>`     | A _namespaced_ keyword like `:my-app/some-req`                         |
+`<?ev-data>`  | An optional _arbitrary edn value_ like `{:data "data"}`                |
+`:ring-req`   | Ring map for Ajax request or WebSocket's initial handshake request     |
+`:?reply-fn`  | Present only when client requested a reply.                            |
 
 #### Summary
 
--- a/example-project/project.clj
+++ b/example-project/project.clj
@@ -21,7 +21,8 @@
    [com.taoensso/timbre       "4.1.1"]
 
    ;;; ---> Choose (uncomment) a supported web server <---
-   [http-kit                  "2.1.19"]
+   [org.clojars.whamtet/dogfort "0.2.0-SNAPSHOT"]
+   ;; [http-kit                  "2.1.19"]
    ;; [org.immutant/web       "2.1.0"] ; v2.1+ recommended
    ;; [nginx-clojure/nginx-clojure-embed "0.4.2"] ; Needs v0.4.2+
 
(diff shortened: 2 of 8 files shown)


whamtet:
Hi, I forgot to mention building.  Here's how.  First install node.  If you're on Linux download rather than apt-get.

    cd sente
    lein install
    cd example-project
    lein npm install #installs npm dependencies
    lein build-once
    node main.js

Visit `https://localhost:5000` for the example project.  At the moment it doesn't have any https support, I can potentially add this if there is demand.

danielcompton:
Hi @whamtet, this looks really interesting, thanks for putting this together. A few initial thoughts

* Could you rebase all of the commits together and give them a descriptive commit message?
* I'll wait for Peter, but I think we may want to get this merged into the main Sente namespace to avoid code duplication and introducing bugs in the future.
* One possible route for avoiding breaking API changes in the main namespace would be to use a Node reader conditional. However this relies on Clojure 1.7 and I don't think Peter's ready to drop support for older Clojure versions. - #149

whamtet:
Ok, renamed the commits.

ptaoussanis:
Hi Matthew, thanks so much for your efforts on this!

The PR in its current form is a little difficult to digest. Some suggestions:
1. Please rebase as a single commit against the `dev` branch
2. Please try avoid unnecessary code formatting changes since these make the diffs harder to parse - I can't tell what changes were actually necessary for functionality and which were just superficial
3. It'd be helpful to have a breakdown of what's been changed + why
4. If at all possible, please try ~stick to the project's code style conventions

3's the most important, for example:
* interfaces.cljx: `#+clj` conds removed since we need these interfaces from ClojureScript
* dogfort.cljs: this is the new adapter
* sente_node.cljs: what is this? What changes were necessary from sente.cljx + why?

Adding an entirely new, duplicate sente ns is a non-starter - but if you can help me understand what needs prompted you to fork the ns, then we could maybe work together to make the fork unnecessary?

Finally, what's the current status of this PR? Is this code working in production? Is it stable? Experimental? Is it waiting on any extra work?

Definitely open to a merge if we can get something self-contained that integrates cleanly with Sente.

Again, much thanks- cheers! :-)

whamtet:
Hi Pete,

Regarding the indentation, what's your editor and settings?  Cider?  What additional config do you use?

Regarding merging node_sente.cljx back into node.cljx the breaking changes are 

----
(defn event-msg? [x] ...) must be split into explicit client and server versions because we can no longer use clj/cljs to distinguish client and server.

unpack* don't want to throw t in cljs when on server
----

all the other changes are stripping of #+clj so that it is also output to cljs.

Do you agree to that?  If so i'll merge it into sente.cljx

Also, it seems to be prod ready.  The example works fine, although not tested in production yet.  I'll do that soon.

whamtet:
Unfortunately there's still a bit of replication when you put everything back into sente.cljx but i'll try it and hopefully it is better

ptaoussanis:
> Unfortunately there's still a bit of replication when you put everything back into sente.cljx

Sorry, I don't follow - could you rephrase this?

ptaoussanis:
> I agree with your comment that we don't want code duplication by creating a new namespace sente_node.cljs.  However a lot of sente.cljx is taken up by the function make-channel-socket!.  If we are going to support node we need to compile both the server and client versions to cljs.  So even if I put it all in sente.cljx most of the duplication is still there.

> I suggest either upgrading to clojure 1.7 with reader conditionals or keeping sente_node.cljs separate until you are willing to do so.  I could also just maintain sente_node.cljs as a separate project.  What do you think?

Hi Matthew,

First step would be identifying the specific changes that were necessary in `sente.cljx` so that I can see about isolating them. That is: I'd like to see a `sente.cljx` with just the diffs necessary for running with node - minus formatting changes.

My hope is that if the necessary changes aren't too extensive, I may be able to mod sente.cljx to support the node adapter without duplicating the entire ns. For example, we can rename some functions like `make-channel-socket-server!`, `make-channel-socket-client!` to uncouple the current notion that client==cljs and server==clj.

Does that make sense?

theasp:
Hello,

I've adapted Sente to support a CLJS server with minimal changes in my dev branch.  I've also added whamtet's dogfort adapter and example project.  Sente seems to work fine, but dogfort crashes if you reload a few times.

https://github.com/theasp/sente/tree/dev

ptaoussanis:
@theasp Hey Andrew, that looks very promising - thank you!

Have my hands full until this weekend, but will definitely try take a look ASAP and get back to you.

> Sente seems to work fine, but dogfort crashes if you reload a few times.

Any chance you could provide some more details on this? How are you reloading? What kind of crash? Is there a specific error message?

whamtet:
Thanks Andrew,

Your wrapper is great.  There are a few more problems at the dogfort end, i'll fix them.

whamtet:
Hi again,

I think i've fixed that bug in dogfort, @theasp Andrew are you still getting those bugs?  If not do a pull request and we can close this one.

Matt

theasp:
@whamtet Your fix to dogfort (in git, not on clojars yet) seems to work.  I'll open a new pull request.

ptaoussanis:
Closing for now; please feel free to reopen if there's renewed interest in getting this into Sente for v1.8.0.

PR #170, opened 2015-10-02
Include client-id docs in the README in the source

kenrestivo:
Rather than pointing to a github issue.


ptaoussanis:
Hey Ken, thanks for this. I'd rather not bloat the size of the README for this - and as Daniel mentioned there'd also need to be some reworking of the content then which I really don't think is urgent.

So I'd prefer to leave the link to the GitHub issue for now. At some time in future may break the README down into a paged Wiki, etc. where it'd make more sense to directly include longer documentation.

Thanks though!

Issue #167, opened 2015-09-29
sente/reconnect doesn't reset backoff-ms attempt number when chsk is closed


danielcompton:
From my testing, it seems like when calling `sente/reconnect` from a closed chsk, the backoff-ms attempt number isn't reset. The use case for this is displaying to the user the number of seconds before reconnecting, but giving them the option to reconnect faster if they want to. Currently, they can click a reconnect button, but if the channel socket is closed then nothing will happen, and Sente will try to reconnect on it's normal schedule.

Is this understanding correct, and is this a bug?

ptaoussanis:
Hey Daniel,

Your understanding's correct - though the behaviour's not _technically_ a bug. `chsk-reconnect!`'s currently documented as: "Drops connection, allows auto-reconnect. Useful for reauthenticating after login/logout."

i.e. `chsk-reconnect!` was originally added as a quick hack as a way of dropping an existing connection to allow it to come back up. Didn't have in mind to allow it to speed up reconnect attempts of already-dropped connections, though that's a totally reasonable expectation from the name.

Would be happy for a PR to adopt the behaviour you're describing (+ update the docstring). Might not be an obvious/easy way to do this btw, so feel free to leave this to me if it's looking like it might be tricky.

Can try take a look at this myself at the weekend. Thanks for the report btw, the current behaviour's definitely surprising.

danielcompton:
I'll have a look tomorrow with clearer eyes, but I think this might be part of the solution (also a note to self):

```clj
(reset! nattempt_ 0)
(.clearInterval js/window @kalive-timer_)
```

ptaoussanis:
> but I think this might be part of the solution

The `nattempt_` looks good, but note that the keep-alive timer and backoff timer's are different. The backoff timer isn't currently accessible for cancelling; it's just an ephemeral thing created in `set-exp-backoff-timeout!`.

So we may need to add a new `backoff-timer_` atom (like `kalive-timer_`) that we _do_ track to allow cancelling.

lfn3:
So I've taken a swing at this, and what I went with is adding the connect! fn to the record and using that directly while clearing the interval timer. This means the timer gets reset but the number of attempts doesn't (which seems desirable?).

Does that sound good to you guys?

danielcompton:
Open a PR and we can take a look :)

ptaoussanis:
Fixed in the SNAPSHOT I'm busy uploading now...

PR #174, opened 2015-10-18
Immediately reconnect when called on with closed socket.

lfn3:
This makes chsk-reconnect!, when called on a closed socket,
attempt to connect immediately. It retains the original behaviour
of incrementing the backoff.

There is one issue with this at present, which is that it can cause
multiple sockets to be opened (if chsk-reconnect! is used repeatedly,
or if the timer fires at exactly the same time, probably.)

Considering adding a mutex "connecting?" to the state atom, but wanted
to check that was sane first.

### DIFF
--- a/src/taoensso/sente.cljx
+++ b/src/taoensso/sente.cljx
@@ -791,6 +791,7 @@
         (when-let [s @socket_] (.close s 3000 "SENTE_RECONNECT")))
       (do
         (.clearInterval js/window @kalive-timer_)
+        (reset! nattempt_ 0)
         (infof "Forced reconnect attempt.")
         (@connect-fn))))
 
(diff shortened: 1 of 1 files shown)


lfn3:
This should fix #167, @danielcompton

danielcompton:
I'd probably expect that forcing a reconnect would restart the backoff count from 0 again. What do you think?

lfn3:
I don't really have strong feelings either way. I've thrown that in.

Any thoughts on the reconnect mutex-y thing? Is there a better way of doing that than adding another bit to the state atom?

ptaoussanis:
Hi there,

Sorry for not taking a look at this sooner. Will definitely get this sorted out for the next release (v1.8.0).

ptaoussanis:
Hi Liam,

Sorry for never getting this merged. Have taken a look and think it'd be better long-term to take a moment to properly refactor the reconnection logic to better support this use case. The current implementation is a big of a kludge.

Your PR definitely helped move this along so again, thank you! Going to keep the original issue (#167) open until I have a fix up.

Cheers :-)

PR #187, opened 2015-12-21
Add carpet as an example

ebellani:
This is the basis for a web application that I've used in a personal project. I though it would be proper to link it here as an example of an illustration of the usage of sente in a more complex scenario.


ptaoussanis:
Awesome, thanks Eduardo!

PR #168, opened 2015-09-30
Allow ClojureScript to act as a server

theasp:
This is the modification to sente to allow ClojureScript to act as a server.  It also introduces the dogfort server adapter and an example project for node.js, both written by Matthew Molly (@whamtet).

The changes to the sente API to allow this is that instead of clj/cljs deciding if sente provides a client or socket, the following was changed:
- `make-channel-socket!` has become `make-channel-socket-server!`
and `make-channel-socket-client!`
- `start-chsk-router!` now has a 3rd argument, `server?`

I am having an issue with the dogfort server adapter and am not sure how you'd like to proceed with it.  Wrapping a few calls in `try` may be all that is needed to fix it.  Perhaps it would be best to strip it out of my pull request and let him submit it, but that would leave us without a working example.  I am going to look into another server.

The following is reproducible by reloading the page several times:
```
phillipsa-lx:/data/phillipsa/projects/sente/node-example-project$ node main.js 
Starting dogfort...
DEBUG [example.my-app-node] - Web server is at http://localhost:5000
DEBUG [example.my-app-node] - Event: [:chsk/uidport-open]
"Unhandled event:" [:chsk/uidport-open]
Broadcasting server>user: {:ws #{:taoensso.sente/nil-uid}, :ajax #{}, :any #{:taoensso.sente/nil-uid}}
DEBUG [example.my-app-node] - Login request: {:user-id "test", :csrf-token "null"}
DEBUG [example.my-app-node] - Login request: {:user-id "test", :csrf-token "null"}
Broadcasting server>user: {:ws #{:taoensso.sente/nil-uid}, :ajax #{:taoensso.sente/nil-uid}, :any #{:taoensso.sente/nil-uid}}

/data/phillipsa/projects/sente/node-example-project/node_modules/ws/lib/WebSocket.js:218
    else throw new Error('not opened');
               ^
Error: not opened
    at WebSocket.send (/data/phillipsa/projects/sente/node-example-project/node_modules/ws/lib/WebSocket.js:218:16)
    at [object Object].taoensso$sente$interfaces$IAsyncNetworkChannel$send_BANG__STAR_$arity$3 (/data/phillipsa/projects/sente/node-example-project/out/taoensso/sente/server_adapters/dogfort.cljs:27:33)
    at taoensso$sente$interfaces$send_BANG__STAR_ (/data/phillipsa/projects/sente/node-example-project/out/taoensso/sente/interfaces.cljs:11:12)
[...]

PR #192, opened 2016-01-29
Split example to client & server

theasp:
As discussed in #168.  Let me know if this is acceptable.

### DIFF
--- a/example-project/project.clj
+++ b/example-project/project.clj
@@ -45,21 +45,23 @@
    [cider/cider-nrepl   "0.10.1"] ; Optional, for use with Emacs
    ]
 
+  :profiles
+  {:clj-server {:source-paths ["src-server"]
+                :main example.server}}
+
   :cljsbuild
   {:builds ; Compiled in parallel
-   [{:id :main
-     :source-paths ["src"]
+   [{:id :cljs-client
+     :source-paths ["src-client"]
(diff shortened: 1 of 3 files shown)


ptaoussanis:
Merging to take a closer look, thanks! Will let you know once I'm ready for the node PR :-)

ptaoussanis:
Hi Andrew, that was helpful - thank you!

Have the updated example project up at https://github.com/ptaoussanis/sente/tree/dev/example-project. **Am ready for the node PR when you are**.

### Re: the node example

After taking a closer look at the split client/server namespaces, it occurs to me that it'd probably be best (simplest) if we just kept the separate client/server namespaces as `client.cljs` and `server.clj` for at least the short term until it becomes clear how many folks are going to be interested in a cljs-based server.

How would you feel if you include the node example as an entry at https://github.com/ptaoussanis/sente/tree/dev#example-projects? You could set up a separate repo for this with a `client.cljs` and `server.cljs` (your preference, of course).

theasp:
I am perfectly fine with hosting a fork of the example.  Hopefully someone will poke me if the API changes.

ptaoussanis:
Not expecting any major changes; Sente's always been pretty stable and I'm pretty happy with the current API. 

If there are ever any unexpected changes, will also try keep backwards compatibility when possible.

PR #196, opened 2016-02-01
Use standard header for CSRF

theasp:
Use the "X-CSRF-Token" header rather than the "csrf" post paramter.
This makes the CSRF token available to ring.anti-forgery using the
default settings as well as csurf when using express.

ring-anti-forgery:
```clojure
(defn- default-request-token [request]
  (or (-> request form-params (get "__anti-forgery-token"))
      (-> request :headers (get "x-csrf-token"))
      (-> request :headers (get "x-xsrf-token"))))
```
csurf:
```javascript
function defaultValue(req) {
  return (req.body && req.body._csrf)
      || (req.query && req.query._csrf)
      || (req.headers['csrf-token'])
      || (req.headers['xsrf-token'])
      || (req.headers['x-csrf-token'])
      || (req.headers['x-xsrf-token']);
}
```

I should point out this would require a change to the ring.anti-forgery middleware setup to account for this, as well as the login ajax call on clients.

### DIFF
--- a/example-project/src/example/client.cljs
+++ b/example-project/src/example/client.cljs
@@ -122,20 +122,21 @@
             ;;; POST request that resets our server-side session. Then we ask
             ;;; our channel socket to reconnect, thereby picking up the new
             ;;; session.
-
-            (sente/ajax-lite "/login"
-              {:method :post
-               :params {:user-id    (str user-id)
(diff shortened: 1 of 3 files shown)


theasp:
I'm working on a method to control this rather than breaking the existing behaviour, please ignore for now.

theasp:
See #198

PR #198, opened 2016-02-02
Make the CSRF token location configurable

theasp:
Rather than having the CSRF location hardcoded, allow specifying
 `:csrf-path` for `make-channel-socket-server!`, which is passed to clients during handshake, describing where a client is to place the CSRF token for posts, to match the configuration of the server.  This allows the server to use the standard middleware without any options for both ring-anti-forgery and csurf when you use a value of `[:headers :X-CSRF-Token]`.  The default is `[:params :csrf-token]` to maintain compatibility with previous versions.

The example server has been upgraded to use `[:headers :X-CSRF-Token]`, and the example client checks to see if there was an error to determine login success or not.

### DIFF
--- a/example-project/src/example/client.cljs
+++ b/example-project/src/example/client.cljs
@@ -122,20 +122,21 @@
             ;;; POST request that resets our server-side session. Then we ask
             ;;; our channel socket to reconnect, thereby picking up the new
             ;;; session.
-
-            (sente/ajax-lite "/login"
-              {:method :post
-               :params {:user-id    (str user-id)
-                        :csrf-token (:csrf-token @chsk-state)}}
-              (fn [ajax-resp]
-                (->output! "Ajax login response: %s" ajax-resp)
-                (let [login-successful? true ; Your logic here
-                      ]
-                  (if-not login-successful?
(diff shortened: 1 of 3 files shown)


ptaoussanis:
Hi Andrew, thanks for your work on this. 

Going to want a little time to consider. May I clarify: the reason for this change is because you don't like having to mod your Ring middleware to suit where Sente sticks its csrf token in requests?

theasp:
No rush.  Yes, that is essentially the reason, for both ring and express, and eliminating differences between the two.  It may make it easier for someone to add sente to an existing app too.  If you decide you don't want it, no big deal.

Thanks

ptaoussanis:
Okay, thanks for the clarification.

ptaoussanis:
Hi Andrew, closing this - will be addressed in the next beta release.

Elected to just duplicate the `:csrf-token` param into the "X-CSRF-Token" header to avoid needing to introduce a new config opt + the client<->server opt sync complexity.

Thanks for bringing this to my attention. Cheers :-)

theasp:
That works too!  Thanks!

Issue #218, opened 2016-04-01
Using sente with Reactjs native breaks when trying to get window location


jlo1:
When trying to bundle a react native project using sente, the call to `make-channel-socket!` breaks. The source code attempts to call `enc/get-window-location`, but I’m getting this error `undefined is not an object (evaluating ‘taoensso.encore.get_window_location.call')`. 

Is the call to get win-location necessary if we specify the path and host ourselves? (Looking at https://github.com/ptaoussanis/sente/blob/a884c90169414ae9f1238dfbe7852693d1505e6c/src/taoensso/sente.cljx#L1061)

ptaoussanis:
Hi there, thanks for the report. This is something I'd like to rectify. Will try take a look when I have some free time, otherwise you might be able to help speed this along with a PR.

Have you confirmed that the window location call is the only problem running on Reactjs native?

Thanks, cheers :-)

jlo1:
Simply adding `[com.taoensso/encore "2.49.0"]` to project dependencies worked fine due to changes in this `encore` commit: https://github.com/ptaoussanis/encore/commit/b869ab3d5d01e313cfd946b379ce7ebc73779f6a

ptaoussanis:
Yeah, that's the commit that did it. Your confirmation was helpful since I couldn't recall if anything else might turn out to be a problem.

How's Sente working with React.js otherwise? Working well?

PR #219, opened 2016-04-04
Only call get-window-location if location is applicable

jlo1:
@ptaoussanis for issue https://github.com/ptaoussanis/sente/issues/218

Tested on a react native project by exporting to a local jar, and it worked fine.

### DIFF
--- a/src/taoensso/sente.cljx
+++ b/src/taoensso/sente.cljx
@@ -1058,7 +1058,7 @@
 
   (let [packer (interfaces/coerce-packer packer)
 
-        win-location (enc/get-window-location)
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Hi Jessica, thanks for the confirmation.

If this is the only change necessary, then no change needs to be made to Sente if you upgrade to the latest encore dependency (currently `[com.taoensso/encore "2.49.0"]`).

So just add `[com.taoensso/encore "2.49.0"]` to your project dependencies before any other `com.taoensso` lib and you should be good.

Cheers! :-)

jlo1:
Thanks. That worked fine!

PR #220, opened 2016-04-08
Little fix to display the message "Channel socket successfully established!"

kajism:
The message "Channel socket successfully established!" was never displayed before this fix. After this place I can safely start to communicate over the channel, so it seems to be pretty important point.


ptaoussanis:
Thanks!

kajism:
My pleasure!

PR #221, opened 2016-04-08
Clear the keep alive timer in `chsk-disconnect!`

theasp:
This prevents multiple timers from piling up.  Not needed in `chsk-reconnect!` as it calls `-chsk-connect!`, which clears it already.


ptaoussanis:
Thanks Andrew, good catch!

Issue #208, opened 2016-03-04
Aleph adapter


ztellman:
I'm creating this to take comments out of the other, mostly unrelated issue.

I started on https://github.com/ztellman/sente, but got stuck because both of the adapters seem to rely on some hybrid websocket/HTTP streaming body mechanism that doesn't seem explicitly specified anywhere.  I can dig into the code, but if you can point me to any documentation for the sente protocol, that would make this easiest for me (and any future adapter authors).

ptaoussanis:
Hey Zach,

> if you can point me to any documentation for the sente protocol, that would make this easiest for me (and any future adapter authors).

Documentation's at https://github.com/ptaoussanis/sente/blob/master/src/taoensso/sente/interfaces.cljx, and there's some example adapters at https://github.com/ptaoussanis/sente/tree/master/src/taoensso/sente/server_adapters

Not sure I follow what you're asking exactly, could you possibly rephrase? Is there something in the docs there that's unclear? Not familiar enough with Aleph to comment directly on your implementation w/o more context, sorry.

> I started on https://github.com/ztellman/sente

Please note that this fork's a little old (1.7.0 vs master at 1.81). Some interface names and docstrings were updated in 1.8 to be clearer which might help?

Looking forward to getting an Aleph adapter in, cheers :-)

Frozenlock:
@ztellman any update on the adapter?

ptaoussanis:
Closing for now, please feel free to reopen if there's any new updates - cheers :-)

ztellman:
That's fair, I'll get to it soon, hopefully.

morgancmartin:
Hi all. I am putting a project together using aleph and sente. I am using the aleph server adapter referenced in the sente docs.

My issue is that when aleph attempts to initialize its websocket connection with the server, an error is thrown by `aleph.http.server/initialize-websocket-handler` stating that "java.lang.ClassCastException: clojure.lang.PersistentHashMap cannot be cast to aleph.http.core.NettyRequest."

On the surface this seems like a pretty straightforward error, but after a couple of days of toying around in the aleph.http, and the sente/aleph adapter, I am ready to check in for some guidance.

I am fairly new to Clojure, otherwise I would continue forging ahead alone. Any help would be much appreciated. Fuller code samples and stacktraces available upon request. And thanks very much @ztellman and @ptaoussanis for the awesome libraries.

morgancmartin:
[...]

PR #225, opened 2016-04-14
Add Aleph adapter

alexandergunnarson:
Continuation of #208 : An adapter for Aleph. Code is heavily borrowed from @ztellman ’s preliminary implementation, but  modified as needed to ensure proper functioning. I’m sure @ztellman can find some holes in it (e.g. with the socket for some reason wanting to close on every message received), but I’ve tested this adapter with the ws and wss sockets I’m using in a remotely deployed test app of mine and it’s been working great so far.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -36,7 +36,8 @@
     {:dependencies
      [[http-kit         "2.2.0-alpha1"]
       [org.immutant/web "2.1.3"]
-      [nginx-clojure    "0.4.4"]]
+      [nginx-clojure    "0.4.4"]
+      [aleph            "0.4.1"]]
      :plugins
      [;;; These must be in :dev, Ref. https://github.com/lynaghk/cljx/issues/47:
       [com.keminglabs/cljx             "0.6.0"]
--- a/src/taoensso/sente.cljx
+++ b/src/taoensso/sente.cljx
@@ -1085,28 +1085,33 @@
                      :<server  (chan recv-buf-or-n)}
 
         ever-opened?_ (atom false)
-        state*        (fn [state]
-                        (if (or (not (:open? state)) @ever-opened?_) state
-                          (do (reset! ever-opened?_ true)
-                              (assoc state :first-open? true))))
+        state*
+        (fn [state]
+          (let [ever-opened? @ever-opened?_]
+            (if (or ever-opened? (not (:open? state)))
+              state
+              (do (reset! ever-opened?_ true)
+                  (assoc state :first-open? true)))))
 
-        ;; TODO map< is deprecated in favour of transducers (but needs Clojure 1.7+)
+        ;; TODO map< is deprecated in favour of transducers
+        ;; (but needs Clojure 1.7+)
 
         public-ch-recv
         (async/merge
           [(:internal private-chs)
-           (async/map< (fn [state] [:chsk/state (state* state)]) (:state private-chs))
+           (async/map<
+             (fn [state] [:chsk/state (state* state)])
+             (:state private-chs))
 
            (let [<server-ch (:<server private-chs)]
              (if wrap-recv-evs?
                (async/map< (fn [ev] [:chsk/recv ev]) <server-ch)
-               (async/map< (fn [ev]
-                             (let [[id ?data] ev]
-                               ;; Server shouldn't send :chsk/ events. As a
-                               ;; matter of hygiene, ensure no :chsk/_ evs are
-                               ;; received over <server-ch
-                               (have? #(not= % "chsk") (namespace id))
-                               ev))
+               (async/map<
+                 (fn [ev]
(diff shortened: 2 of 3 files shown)


ptaoussanis:
Hi Alex, thanks for this.

> (e.g. with the socket for some reason wanting to close on every message received)

Sorry just to clarify: that's a hypothetical example, or this actually happens with the implementation in the PR?

alexandergunnarson:
Hey, no problem at all! And that's what actually happened when I didn't comment out the body of `when close-after-send?`. Since it's commented out in the PR, it doesn't happen. The downside of this is that it requires the client to explicitly close a socket (no implicit closing via `close-after-send?`) unless the client gets disconnected or navigates away from the page, in which case of course the socket will be closed anyway. After some playing around with it, it seemed apparent to me that the solution in the PR was the only viable one. I'd be happy for you or @ztellman to prove me wrong, of course.

ztellman:
I'll take a look at the implementation over the next few days.

alexandergunnarson:
Sounds great. Thanks! I probably only changed a few lines of code from your implementation to make it work — the main thing was to fix some arity issues of some anonymous function or other, which was a pain to track down. As I mentioned, this implementation works perfectly for my use case but there may be some who rely heavily on `close-after-send?` — not sure why that's breaking so that's probably the main focus for now.

ptaoussanis:
> As I mentioned, this implementation works perfectly for my use case but there may be some who rely heavily on close-after-send? — not sure why that's breaking so that's probably the main focus for now.

Just have my hands full atm, so haven't looked at any of this - but just want to note that `close-after-send?` support is not optional (it's used for Ajax connections).

alexandergunnarson:
Ah, well I'll look into why that part isn't working then. My use case, of course, is not everyone else's.

ztellman:
I just realized I completely forgot about this.  I guess I don't understand the exact nature of the issue with `close-after-send?`.  @alexandergunnarson, what does it mean that it closes "every time, even with a callback"?

alexandergunnarson:
@ztellman The callback I'm referring to is the callback passed as the third argument to `chsk-send!`.
For instance, from [the example Sente provides](https://github.com/ptaoussanis/sente/blob/master/example-project/src/example/client.cljs):

With callback:
```
(chsk-send! [:example/button2 {:had-a-callback? "indeed"}] 5000
  (fn [cb-reply] (->output! "Callback reply: %s" cb-reply)))
```

Without callback:
```
(chsk-send! [:example/button1 {:had-a-callback? "nope"}])
```

According to the internal logic of Sente (which I can hunt down to give you more context, or @ptaoussanis can tell you more quickly and precisely), if such a callback is not passed, the websocket closes automatically; otherwise, it is left open, awaiting a callback response. This is triggered by `close-after-send?` (the third argument to `-sch-send!`) being passed as true. Ignoring the flag `close-after-send?` flag and keeping the websocket open is normally totally fine (e.g. for my use case), but as @ptaoussanis noted, "close-after-send? support is unfortunately not optional (it's used for Ajax connections)".

Let me know if you want/need further context.

ptaoussanis:
Thanks Alex!

> According to the internal logic of Sente (which I can hunt down to give you more context, or @ptaoussanis can tell you more quickly and precisely), if such a callback is not passed, the websocket closes automatically; otherwise, it is left open, awaiting a callback response. 

Only had a moment to skim this issue, but just want to clarify quickly in case it's useful that this doesn't actually have anything to do with callbacks.

We're talking about Aleph's implementation of [this](https://github.com/ptaoussanis/sente/blob/e1903c0173693d1aa3a0bc35f915ec4186d87a96/src/taoensso/sente/interfaces.cljx#L27) server-side protocol function. It's used to send data to connected clients (might be a WebSocket client or an Ajax client).

For WebSockets: `close-after-send?` will always be false. I.e. we never need/want WebSockets to close after a send.

For Ajax conns: `close-after-send?` will always true. I.e. we want regular HTTP req->resp conns to close after the first response (we're not using chunked transfer encoding, etc.).

Please let me know if there's anything else I can help clarify, will do my best to respond ASAP.

alexandergunnarson:
Thanks @ptaoussanis for your clarification/correction. I haven't been working on the WebSocket-related project I mentioned in probably 3 weeks, so apparently my memory is becoming inaccurate. Either way, callback or not, I remember there definitely being client- and server-side WebSocket support when I was using the Aleph adapter, so there was never an Ajax connection; however, despite this, `close-after-send?` was always true. Thus I simply disabled it in `-sch-send!`. This would sound like something orthogonal to Aleph were it not for the fact that this problem never occurred with e.g. Immutant. 

I will play around with this more, but it might be another few weeks before I get back to that project — sorry.

ptaoussanis:
Closing this to focus on #236

PR #228, opened 2016-04-20
Removed unused imports

Rovanion:
I feel like I must begin by asking for forgiveness in case I'm simply wasting your time. But I was wrapping my head around the example code and found some imports which weren't being used in the code so. As a newcomer to the Clojure world I simply tried removing them to see what happened and it seems to me like nothing at all happened.

As far as I can see all the features of the demo is still working after my changes. So perhaps removing the unused imports will help some newcomer in the future by not making them trying to figure out what they do.

But do feel entirely free to discard this PR if there is anything wrong with it.

### DIFF
--- a/example-project/src/example/client.cljs
+++ b/example-project/src/example/client.cljs
@@ -4,10 +4,9 @@
 
   (:require
    [clojure.string  :as str]
-   [cljs.core.async :as async  :refer (<! >! put! chan)]
-   [taoensso.encore :as encore :refer ()]
-   [taoensso.timbre :as timbre :refer-macros (tracef debugf infof warnf errorf)]
-   [taoensso.sente  :as sente  :refer (cb-success?)]
+   [taoensso.encore :as encore]
+   [taoensso.timbre :as timbre]
+   [taoensso.sente  :as sente]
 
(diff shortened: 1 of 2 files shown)


xcthulhu:
This looks good.

I'll hopefully have some time this week to update my unit testing system so I can confirm these changes are okay.

ptaoussanis:
Hi Rovanion,

Thanks for this (+ no forgiveness necessary). There are pros + cons to removing these imports. They're technically unnecessary (as you've seen), but they're there since they might be useful / are commonly used (e.g. core.async, some basic logging utils, etc.).

The server-side `encore` import could go, that's mostly there out of habit. It is used in the client ns though, so part of me is inclined to keep that too just to be consistent (so the same utils are available in both namespaces).

Going to decline for now, but might take a look at documenting optional imports next time I'm in the example code.

Appreciate your time+input on this, cheers :-)

Rovanion:
Yeah, it's an opinionated PR so I'm entirely fine with you disagreeing with it. And I highly appreciate the time you've spent developing Sente and subsequently the time you're spending supporting it right now. Cheers!

PR #236, opened 2016-06-10
working aleph server adapter (i think?)

sorenmacbeth:
Hello,

As far as I can tell this works, provided that when you create your Ring middleware, you ensure that you add the `{:websocket? true}` key to the request when appropriate. I'm not sure if this is acceptable, or if I can do it directly in the server adapter, so some guidance here would be helpful.

I've test that websocket clients work under the conditions mentioned above. I haven't directly tested that ajax polling works though.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -37,7 +37,8 @@
     {:dependencies
      [[http-kit         "2.2.0-alpha2"]
       [org.immutant/web "2.1.4"]
-      [nginx-clojure    "0.4.4"]]
+      [nginx-clojure    "0.4.4"]
+      [aleph "0.4.1"]]
      :plugins
      [;;; These must be in :dev, Ref. https://github.com/lynaghk/cljx/issues/47:
       [com.keminglabs/cljx             "0.6.0"]
--- a/src/taoensso/sente/server_adapters/aleph.clj
+++ b/src/taoensso/sente/server_adapters/aleph.clj
@@ -1,0 +1,37 @@
+(ns taoensso.sente.server-adapters.aleph
+  (:require
+    [taoensso.sente.interfaces :as i]
+    [aleph.http :as http]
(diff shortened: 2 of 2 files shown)


sorenmacbeth:
afaik this isn't suffering from the problem in #225

ptaoussanis:
Thanks Soren, will try test this w/in the next few days.

> provided that when you create your Ring middleware, you ensure that you add the {:websocket? true} key to the request when appropriate

Sorry, could you elaborate on this a bit? Ring middleware is required that adds `{:websocket? true}` to the Ring request when `websocket-request?` is true?

sorenmacbeth:
I had to add this middleware to my application's sente aleph server implementation like this:

```clojure
(defn websocket?
  [req]
  (if (= "websocket" (get-in req [:headers "upgrade"]))
    (assoc req :websocket? true)
    req))
```

then in my sente route I did:

```clojure
(GET  "/chsk" req
        (debugf "/chsk got: %s" req)
        (ring-ajax-get-or-ws-handshake (websocket? req)))
```

ptaoussanis:
Oh, as I remember - may I also ask that you add the necessary Aleph stuff to the reference example project at:
https://github.com/ptaoussanis/sente/blob/master/example-project/project.clj
and
https://github.com/ptaoussanis/sente/blob/master/example-project/src/example/server.clj?

Please ignore the `:websocket?` middleware for now, I'll handle that. Thanks!

ptaoussanis:
Okay, managed to take a look at this today. The `sch-send!` implementation needed some tweaking, otherwise this seems to be working okay.

Btw think you actually identified the problem with #225 here; looks like it was an issue with the missing middleware.

Much thanks, merged!

sorenmacbeth:
woo hoo!

PR #243, opened 2016-07-10
Allow nodejs clients

DaveWM:
This is a fix for issue #209. I've defined a `global` var (`js/window` on browsers, or `js/global` on node), and replaced all references to `js/window` with this var. 

To get websockets working on node, I added the npm [websocket](https://www.npmjs.com/package/websocket) library as a dependency. I couldn't see this (or anything similar) in clojars, so I had to add the [lein-npm](https://github.com/RyanMcG/lein-npm) plugin and import the library directly from npm.

I've also added a nodejs example to the example project, and instructions on how to run it.


project.clj
```diff
@@ -48,7 +48,9 @@
       [lein-ancient                    "0.6.10"]
       ;; [com.cemerick/austin          "0.1.4"]
       [com.cemerick/clojurescript.test "0.3.3"]
-      [lein-codox                      "0.9.5"]]}]}
+      [lein-codox                      "0.9.5"]
+      [lein-npm                        "0.6.2"]]}]}
+  :npm {:dependencies [[websocket "1.0.23"]]}
```
danielcompton: We need to be careful that this NPM dependency doesn't leak downstream. I'm pretty sure it won't, but worth checking.

src/taoensso/sente.cljx
```diff
@@ -928,8 +943,10 @@
               false))))))
 
   (-chsk-connect! [chsk]
-    (when-let [WebSocket (or (enc/oget js/window "WebSocket")
-                             (enc/oget js/window "MozWebSocket"))]
+    (when-let [WebSocket (if is-node?
+                           node-websocket
```
danielcompton: Can we include `node-websocket` in the `or` conditional below? It'll be nil if not running on node.

example-project/src/example/node.cljs
```diff
@@ -0,0 +1,35 @@
+(ns example.node
```
danielcompton: Would be good to add a comment here pointing out this is a node JS client.

src/taoensso/sente.cljx
```diff
@@ -81,6 +81,21 @@
   (:require-macros
    [cljs.core.async.macros :as asyncm :refer (go go-loop)]))
 
+; Note: requiring goog.global would be better, but currently breaks the cljs build: http://dev.clojure.org/jira/browse/CLJS-1677?page=com.atlassian.streams.streams-jira-plugin:activity-stream-issue-tab
```
danielcompton: You don't need to require `goog` to get global. I'm pretty sure that `goog.global` is automatically in scope without a require.

ptaoussanis:
@DaveWM Hi David, thanks for this!

@danielcompton Will wait on your go-ahead to give this a detailed look for merging, thanks!

@DaveWM One request: I'd prefer if we not add any additional examples directly to the Sente repo since that increases the maintenance burden. Examples can be added as external repos with a link at the top of the list at https://github.com/ptaoussanis/sente#example-projects.

Cheers :-)

project.clj
DaveWM: I've checked this in the example project, I can't see the npm dependency in there.

DaveWM:
@danielcompton I've made the changes you suggested. Didn't know that you don't have to require the `goog` namespace, thanks for the tip!

@ptaoussanis I've reverted the changes I made to the example project. If you need to test my changes on nodejs, just check out commit b7fcbaf and run the nodejs example.

Thanks guys, let me know if you need anything else from me :)

theasp:
@DaveWM, if you'd like, I'll accept a PR for the node client in my sente nodejs example repository once Peter releases a version or snapshot with your changes:
https://github.com/theasp/sente-nodejs-example/

project.clj
danielcompton: This NPM dep does leak downstream: https://github.com/RyanMcG/lein-npm/pull/13. To test this, add the `lein-npm` plugin to the example project and run `lein deps`. You'll see `lein-npm` installs `websocket`. `lein-npm` looks into a projects dependencies to see if any of them have any `:npm` dependencies to install too. I'd probably be inclined to remove this dependency entirely here and just make a note of it as being required for Node users to add it to their own projects?

src/taoensso/sente.cljx
```diff
@@ -81,6 +81,16 @@
   (:require-macros
    [cljs.core.async.macros :as asyncm :refer (go go-loop)]))
 
+#+cljs
+(def is-node? (nil? enc/js-?win))
```
danielcompton: Not sure if this is the best detection for `is-node?`, perhaps https://github.com/iliakan/detect-node/blob/master/index.js or another method would be safer?

src/taoensso/sente.cljx
```diff
@@ -81,6 +81,16 @@
...
+
+#+cljs
+(def node-websocket
+  (when (and is-node? (exists? js/require))
+        (-> (js/require "websocket")
+            (aget "w3cwebsocket"))))
+
+
```
danielcompton: Extra space not needed here.

danielcompton:
@ptaoussanis looks pretty good to me, pending removing the npm deps and checking the `is-node?` def is the best option.

src/taoensso/sente.cljx
ptaoussanis: Might `(= *target* "nodejs")` be appropriate for this? Zero experience with node, so not sure.

DaveWM: Does `*target*` come from the `:target` compiler option? If it does, that seems like the way to go. Otherwise, I think checking for `js/window` is the standard way to detect nodejs, but I can add an extra check for the `process` global var if you want.

ptaoussanis: That's correct, here's the docstring:

"Var bound to the name value of the compiler build :target option.
  For example, if the compiler build :target is :nodejs, *target* will be bound
  to \"nodejs\". *target* is a Google Closure define and can be set by compiler
  :closure-defines option."

DaveWM:
I've updated the `is-node?` function, and also removed the `lein-npm` plugin and added a note to the readme.

danielcompton:
LGTM

src/taoensso/sente.cljx
```diff
@@ -928,8 +937,9 @@
               false))))))
 
   (-chsk-connect! [chsk]
-    (when-let [WebSocket (or (enc/oget js/window "WebSocket")
-                             (enc/oget js/window "MozWebSocket"))]
+    (when-let [WebSocket (or (enc/oget goog/global "WebSocket")
+                             (enc/oget goog/global "MozWebSocket")
+                             node-websocket)]
```
ptaoussanis: Is it possible for `goog/global` to have "WebSocket" or "MozWebSocket" when `(= *target* nodejs)`?

ptaoussanis:
@DaveWM Merged, thank you!

Please note that I've removed the README addition; would prefer any client/server-specific docs to go with their example repo, Ref. #244

PR #246, opened 2016-07-15
Add :protocol parameter to make-channel-socket-client!.

tiensonqin:
Since react native don't support window.location.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -1310,6 +1310,8 @@
      Common options:
        :type           ; e/o #{:auto :ws :ajax}. You'll usually want the default (:auto).
        :host           ; Server host (defaults to current page's host).
+       :protocol       ; Custom protocol, #{"https:" "http:"}.
+                       ; React native don't have window.location.
        :params         ; Map of any params to incl. in chsk Ring requests (handy
                        ; for application-level auth, etc.).
        :packer         ; :edn (default), or an IPacker implementation.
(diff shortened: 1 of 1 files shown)


danielcompton:
It's probably better to pass the websocket protocol here ws/wss, rather than http/https

tiensonqin:
@danielcompton Yeah, thanks for advice, I agree with you.

src/taoensso/sente.cljc
```diff
@@ -1310,14 +1310,16 @@
      Common options:
        :type           ; e/o #{:auto :ws :ajax}. You'll usually want the default (:auto).
        :host           ; Server host (defaults to current page's host).
+       :protocol       ; Custom protocol, #{:wss :ws}.
+                       ; React native don't have window.location.
```
danielcompton: React native doesn't have window.location, so we can't infer the protocol.

src/taoensso/sente.cljc
```diff
@@ -1339,15 +1341,18 @@
      (let [packer (coerce-packer packer)
 
            win-loc (enc/get-win-loc)
-           win-protocol  (:protocol win-loc)
+           protocol (case (keyword protocol)
+                      :wss "https:"
+                      :ws  "http:"
+                      (:protocol win-loc))
```
danielcompton: If we're going to allow the user to specify a protocol, then that should go first. Probably an `or` here would be appropriate?

src/taoensso/sente.cljc
```diff
@@ -1339,15 +1341,18 @@
      (let [packer (coerce-packer packer)
 
            win-loc (enc/get-win-loc)
-           win-protocol  (:protocol win-loc)
+           protocol (case (keyword protocol)
```
danielcompton: We need to get the protocol from `win-loc` if it's not provided.

danielcompton: Probably good to add an assert if `win-loc` is nil and `protocol` not provided, and in the error explain that `protocol` must be provided.

danielcompton:
Hmm, don't have time to look at it this second, but how does the Node client do this? It too won't have a `window.location`

ptaoussanis:
Merged, thanks!

PR #252, opened 2016-08-04
Log how much time until next reconnect

danielcompton:
This should make the logs more informative for knowing when the next connection will be made.


ptaoussanis:
Merged, thanks.

PR #256, opened 2016-08-07
Add data15-blackjack to Example Projects

tfoldi:
This one is from last year, but might be useful for newcomers


ptaoussanis:
Thanks!

PR #258, opened 2016-08-09
Warn about lost handle when we expected to have a handle

danielcompton:
If we intentionally disconnect a Sente chsk, previously we would have
gotten a warning "Aborting reconnect attempt (lost handle)" when the
retry handler ran. After this commit, this warning only shows when we
expect we should have a retry handle.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -983,7 +983,9 @@
                connect-fn
                (fn connect-fn []
                  (if-not (have-handle?)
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Hi Daniel, thanks for this. Think it'd be preferable to just remove the warning altogether (it's not interesting or user-actionable). Have pushed a change for this. Cheers! :-)

danielcompton:
Also a good option :)

PR #263, opened 2016-09-12
Add :next-reconnect key to chsk state

danielcompton:
It is often desirable to know when the chsk will next attempt to reconnect, to show the user some UI for this. This patch adds a key :next-reconnect which adds the udt of the next reconnect time to the
state map when the chsk is offline, and removes it when it is online, or stopped.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -55,6 +55,7 @@
     :last-close     - ?{:udt _ :reason _}, with reason e/o
                         #{nil :requested-disconnect :requested-reconnect
                          :downgrading-ws-to-ajax :unexpected}
+    :next-reconnect - ?udt of next reconnect attempt
 
   Notable implementation details:
     * core.async is used liberally where brute-force core.async allows for
@@ -828,11 +829,12 @@
                    :requested-reconnect
                    :downgrading-ws-to-ajax
                    :unexpected}] reason)
-     (if (or (:open? state) (not= reason :unexpected))
-       (assoc state
-         :open? false
-         :last-close {:udt (enc/now-udt) :reason reason})
-       state)))
+     (let [state (dissoc state :next-reconnect)]
+       (if (or (:open? state) (not= reason :unexpected))
+         (assoc state
+           :open? false
+           :last-close {:udt (enc/now-udt) :reason reason})
+         state))))
 
 #?(:cljs
    (defn- cb-chan-as-fn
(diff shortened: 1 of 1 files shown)


danielcompton:
@ptaoussanis have you had a chance to look at this?

ptaoussanis:
Sorry Daniel, just been under a lot of stress at work. Will take a look now.

src/taoensso/sente.cljc
```diff
@@ -828,11 +829,12 @@
                    :requested-reconnect
                    :downgrading-ws-to-ajax
                    :unexpected}] reason)
-     (if (or (:open? state) (not= reason :unexpected))
-       (assoc state
-         :open? false
-         :last-close {:udt (enc/now-udt) :reason reason})
-       state)))
+     (let [state (dissoc state :next-reconnect)]
```
ptaoussanis: Might be missing something, but is there a reason you'd want to drop the next reconnect info here?

To be clear, could you possibly briefly describe the intended behaviour. For example:

1. We set `:next-reconnect` any time we start an auto-reconnect timer
2. Under what circumstances would we want to remove this info?

My initial hunch would be to say on any reconnect, or on any :last-close reason change? Haven't had an opportunity to give it much though though.

Is it possible you wanted this dissoc 2 lines down, as part of the `then` branch?

src/taoensso/sente.cljc
```diff
@@ -1043,7 +1048,9 @@
                                  (or
                                    (when (handshake? clj)
                                      (receive-handshake! :ws chsk clj)
-                                     (reset! retry-count_ 0))
+                                     (reset! retry-count_ 0)
+                                     (swap-chsk-state! chsk
```
ptaoussanis: Please note that `swap-chsk-state!` can return falsey, so this form needs to be followed up with a `true` to prevent possible interference with the enclosing `or`

src/taoensso/sente.cljc
```diff
@@ -55,6 +55,7 @@
     :last-close     - ?{:udt _ :reason _}, with reason e/o
                         #{nil :requested-disconnect :requested-reconnect
                          :downgrading-ws-to-ajax :unexpected}
+    :next-reconnect - ?udt of next reconnect attempt
```
ptaoussanis: Would prefer we name this `:udt-next-reconnect` and describe it as "Approximate udt of next scheduled auto-reconnect attempt".

danielcompton: Done

src/taoensso/sente.cljc
danielcompton: Yep, you're right. I didn't click that this was called under unexpected disconnections. Have moved this down below.

src/taoensso/sente.cljc
danielcompton: Good catch, didn't realise that. Fixed.

danielcompton:
Not sure what happened with the review comments, I made another commit so that GitHub wouldn't lose them...

ptaoussanis:
Before merging, could I please ask for a confirmation of the precise intended/desired behaviour re: when we want to remove :next-reconnect info.

For example: it's currently being removed on successful handshake ajax handshake but not on other successful polling requests.

A clear statement of the intended/desired behaviour would help me check that the logic is correct, thanks.

ptaoussanis:
Unfortunately won't have an opportunity to spend more time on this atm, so just took a quick stab at what I expect your desired behaviour was with https://github.com/ptaoussanis/sente/commit/9e24f459844bc2e7021138cedc041b7881b2237a.

Please let me know if that's acceptable and behaves as you intend in your testing? Thanks!

danielcompton:
> Before merging, could I please ask for a confirmation of the precise intended/desired behaviour re: when we want to remove :next-reconnect info.

Problem statement: If a user chsk is retrying to connect/reconnect, we want to be able to display to them when the next automatic retry will be.

I'll take a look at this tomorrow morning with a clearer head to let you know if this fits. Thanks again!

danielcompton:
Have tested these changes, and what you've done looks good. Thanks!

Issue #262, opened 2016-09-09
How to emulate sente's protocol?


arichiardi:
Hello @ptaoussanis, I was trying to benchmark my sente app and I was wondering if I can somehow emulate the messages sent back and forth. I guess I would need to know the protocol for the handshake at the very basic. Is there a place I can find that (also in the code, and I can contribute it back in some format)?

An alternative would be for me to override sente, disabling it altogether and use the immutant websocket so that I can at least have some kind of "comparable" benchmark.

I don't even know if I am approaching the problem in the correct way and I open to suggestions.

Thanks a lot for your work (it is nice to write this in every issue I think :smile:),
Andrea.

ptaoussanis:
Hi Andrea,

> Thanks a lot for your work (it is nice to write this in every issue I think 😄),

Heh heh, thank you :-)

> I was trying to benchmark my sente app and I was wondering if I can somehow emulate the messages sent back and forth.

Hmm, how clear are you on precisely what you want to benchmark?

Unless you have something else very specific in mind, a decent general idea might be to just write a util to execute real, synthetic Sente events (e.g. `[:my-app/fetch-booking-info {:search-criteria _])`, using the normal Sente API. Just generate events from a script instead of from user input.

This'd be the easiest thing to implement and give you a high-level overview of ~realistic app performance as users would see it.

Your performance numbers would include:
* Application performance (time taken to respond to app-level channel socket requests)
* Sente performance (any overhead that Sente might introduce over your web server)
* Web server performance (e.g. WebSocket handling, etc.)

If you setup your benchmarks to run over the net (perhaps even through a proxy), you could also cover things like your server + CDN latency, etc.

That'd be a decent place to start. If the numbers are too slow, you can then do further tests to figure out which component is bottlenecking. In almost all cases, times will be dominated by:

1) The application if it's hitting the db and not being cached well
2) Network latency

Web server benchmarks are rarely interesting from a performance perspective IMO, esp. for real applications. The cost to generate a response usually so-far outweighs the cost of the web server responding that it's rarely even worth thinking about. Depends on the use case though, of course.

Hope some of that was helpful?

arichiardi:
Oh definitely helpful, thanks!

I was thinking exactly that: to reproduce the protocol and the `:chsk` messages...

A thing I do is for example to save the uuids I receive during the first handshake in some atom. So I need to spec the messages first. I will report back my digging.

ptaoussanis:
[...]

PR #264, opened 2016-09-14
Add how to debug Sente websocket in README

arichiardi:
Follow up on https://github.com/ptaoussanis/sente/issues/262.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,32 @@ Most of Sente's state is held internally to each channel socket (the map returne
 
 The only thing you _may_[1] want to do on component shutdown is stop any router loops that you've created to dispatch events to handlers. The client/server side `start-chsk-router!` fns both return a `(fn stop [])` that you can call to do this.
 
+#### How do I debug Sente websocket packets on the wire?
+
+You can directly inspect the format of sente packets by using `Wireshark` or similar tools. Assuming `sente` does not degrade to `AJAX`, the initial message, the one that ["upgrades"](https://tools.ietf.org/html/rfc6455#section-1.2) to Websocket, ALWAYS includes the `client-id` as __query parameter__:
+
+```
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Thanks, will take a closer look at this next time I'm on Sente!

Cheers :-)

arichiardi:
I will need to add the wrapping to this PR, which I have just recently figurared out and successfully emulated. I was even wondering if it would make sense to extract the part of the `pack` function that adds `+/-` and does the wrapping...but ok we'll discuss that when you'll have time, I got my workaround :smi

arichiardi:
Other section added, I'll leave it here for a double/triple/quadruple check ;)

ptaoussanis:
Hi Andrea-

Sorry for taking so long to get to this, wanted an opportunity to think about it some more.

Appreciate the effort you put into the PR, but I'll be honest - am still hesitant to document implementation details or to encourage folks to depend on them, even for benchmarking.

As I mentioned at https://github.com/ptaoussanis/sente/issues/262#issuecomment-246158984, I'm still not very clear on why one would want to work at a protocol level. Unless there's a strong reason to, would really recommend avoiding it if possible since these details are likely to break unexpectedly one day and without notice.

Am also keen to keep the README as small as possible (it's already on the long side, I think).

As a compromise, have moved your notes to https://gist.github.com/ptaoussanis/848d41c33eb8b4512193ba5e508f961b which we can link to from the FAQ for folks that are interested.

A better long-term solution might be to consider setting up a user Wiki with a proper table of contents, etc. Then we can go into more detail on subjects like this without becoming overwhelming.

Anyway, hope the present solution might be reasonable in the meantime?

ptaoussanis:
Oh, in fact just setup a wiki: https://github.com/ptaoussanis/sente/wiki/Debugging-and-benchmarking-at-the-protocol-level

arichiardi:
Sounds very reasonable, thanks again, agree that this kind of a niche, but it is also good that we now have it :wink:

PR #301, opened 2017-05-07
Bump core.async and aleph versions for new specs in clojure 1.9

pkova:
Sente does not work with the new specs since it uses an old version of core.async which has a bad :refer-clojure. Fixed in [this commit.](https://github.com/clojure/core.async/commit/2f87bc7c7d10cb7b0baafc86e08489d58fa87424)

Aleph version 0.4.1 has a malformed defn so I bumped that to 0.4.3 as well.

See CLJ-2062 for details.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,7 @@
 
   :dependencies
   [[org.clojure/clojure      "1.7.0"]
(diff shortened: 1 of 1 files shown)


danielcompton:
This seems pretty unobjectionable.

PR #317, opened 2017-12-12
Checking for an incoming nil-event in router

boldaslove156:
Checking for an incoming `nil` event in `start-chsk-router!` so that it plays nice with some specfic library, and I think in general checking `nil` in a  `go-loop` is a good idea


ptaoussanis:
Merged manually, thanks! :+1:

PR #323, opened 2018-06-01
Static file anti-forgery is not in session, but is in request

timothypratley:
See upstream ring issue resolved:
https://github.com/ring-clojure/ring-defaults/issues/23
This change looks for the key in the request which was added in
ring-anti-forgery 1.3.2, which is in ring defaults 0.3.2
If you are affected by this issue, make sure you are using Sente 0.13.0+
and ring defaults 0.3.2+


ptaoussanis:
@timothypratley Merging now manually, thanks so much Timothy! Sorry for the long delay on this.

Issue #326, opened 2018-08-22
:auto connection not downgraded on initial ws connection failure


michaelcameron:
We have a failure mode in dev/qa environments with a misconfigured reverse proxy between client and server that doesn't understand or abide by the initial websocket connection attempt and responds with 200 instead of 100. Sente continues to try to reconnect the websocket and never seems to failover to the ajax connection. I would have expected it to fail over. It looks like the auto->ajax downgrade watch is looking for never-opened websocket connection with a `:last-error` in the state map (https://github.com/ptaoussanis/sente/blob/master/src/taoensso/sente.cljc#L1347), but it doesn't look like anywhere else in the code is setting `:last-error`. There are a number of references to `:last-ws-error` and that is in the state map in this case. Is that what the downgrade watch should be looking at, or is it not expected to failover to the ajax connection in this failure mode?

ptaoussanis:
Fixed with release of `[com.taoensso/sente "1.13.1"]`, thanks Michael!

PR #327, opened 2018-08-22
Fix ajax downgrade logic

michaelcameron:
Re: https://github.com/ptaoussanis/sente/issues/326, I made this change locally and it falls back to the ajax connection upon the initial websocket failure. It seems like the code should be looking `:last-ws-error` used elsewhere in code instead of `:last-error` which does not appear to be used.


ptaoussanis:
@michaelcameron Thanks for catching + fixing this Michael! Merged manually.
@timothypratley Thanks for reviewing!

`[com.taoensso/sente "1.13.1"]` on Clojars now 👍

Issue #322, opened 2018-05-11
Assert failed: (every? valid-attr? attrs)


timothypratley:
Sometimes when adding a dependency, you get an older version of `ring`

## Repro

Edit the example project of sente by adding bidi as a dependency:

    [bidi "2.1.3"]

Now when you run the example with `lein start` you get `Assert failed: (every? valid-attr? attrs)`.

`lein deps :tree` reveals that bidi brought in an older version of `ring`.

## Workaround

Prevent the older version from being brought in:

    [bidi "2.1.3" :exclusions [ring/ring-core]]

It would be great if Sente could somehow raise a better exception, but I don't see a way to do that... I'm filing this issue just so that other people searching for it can find the cause.

PR #333, opened 2018-10-18
Exclude ring from immutant dependency

tooke7:
Trying to use Immutant in the example project makes the server crash with the same error message described in #322; this fixes it.


ptaoussanis:
Thanks Jacob, sorry for the long delay to merge!

PR #337, opened 2019-03-08
Fix incorrect value (only udt) swapped into conns_

osbert:
The value stored per client session is

  `[server-channel-socket, timestamp]`

but in this particular instance only the timestamp was swapped in.


ptaoussanis:
Thanks Osbert, merging manually now! 👍

PR #341, opened 2019-04-05
Make cljsbuild output-to resources/public/main.js directly

shaolang:
Before this change, the example-project couldn't run on Windows-based machines. This change makes it possible to run on any.


ptaoussanis:
Thanks shaolang, merged manually 👍

Issue #247, opened 2016-07-15
Unable to resolve module websocket in react native release mode


tiensonqin:
Sente 1.9.0-beta2 works fine, 1.10.0-SNAPSHOT failed.

Error log:
UnableToResolveError: Unable to resolve module websocket from /Users/tienson/codes/index.ios.js: Unable to find this module in its module map or any of the node_modules directories under /Users/node_modules/websocket and its parent directories.

danielcompton:
The node client requires the `websocket` module to be installed. This should probably be a bit more visible somewhere, either in the Sente README, or in https://github.com/theasp/sente-nodejs-example

theasp:
I've added a note to the nodejs example about requiring "websocket".  I'm not sure how this impacts react native though.

ptaoussanis:
This must be due to #243.

@tiensonqin I'm not familiar with React Native, a few questions:

- You must be building with `cljs/*target*` set to `"nodejs"`, is that correct?
- What's the relationship between React Native and node? It can use node? It must use node?
- Any idea if React Native was working correctly over WebSockets with Sente 1.9.0-beta2?

Thanks!

ptaoussanis:
Have updated 1.10.0-SNAPSHOT. I believe this should resolve the error.

@tiensonqin 2 more questions if I may?:

1. Could you confirm that your error's gone with the updated SNAPSHOT?
2. Do WebSockets work with this SNAPSHOT even if you don't have the websockets npm module installed?

tiensonqin:
@ptaoussanis Sorry about the delay.
1. I still have the error after updated to the latest commit, 
    as a workaround I comment out the `?node-npm-websocket`, don't have time for now, sorry!
2. As I know it, react native don't use `npm websocket`,  many npm modules even node core modules can't be used in react native, since they are different platform.
    react active just make websocket web browser compatible by bridgeing `obj-c` or `java`. You don't need to install `npm websocket` in react native.

ptaoussanis:
@tiensonqin Thanks for the follow-up.

---

Can anyone with Node experience perhaps point me to the docs for `js/require`? Had assumed it'd just throw if the requested library wasn't available, but perhaps that's not the case.

ptaoussanis:
@DaveWM Alternatively: do we really need to call `(js/require "websocket")` ourselves? Is that a common idiom for Node? Could we not instead just check if the library's already been loaded and leave Sente users to load the lib how/where/when they want?

[...]

PR #353, opened 2019-11-04
Revert host overriding functionality

g7s:
A previous PR broke the host overriding functionality. This commit fixes it.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -1501,9 +1501,8 @@
                 (f path win-loc :ajax)]
 
                (let [protocol (or protocol (:protocol win-loc) :http)
-                     host     (if port
-                                (str (:hostname win-loc) ":" port)
-                                (or host (:host win-loc)))]
+                     host     (cond-> (or host (:host win-loc))
+                                port (str/replace-first #"(:\d+)?$" (str ":" port)))]
                  [(get-chsk-url protocol host path :ws)
                   (get-chsk-url protocol host path :ajax)])))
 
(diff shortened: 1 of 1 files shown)


Deraen:
@ptaoussanis This is important. Sente is currently unusable e.g. in React-native apps, as window location doesn't mean anything.

But I think this fix should also support case where both `:host` and `:port` are set? Or at least the docstring should be updated to reflect that `:port` is only useful when connecting to current host, but with different port.

g7s:
@Deraen You can provide the port as part of the `:host` and it works. On the other hand it should support `:hostname` and `:port` options (not only port) because as you pointed out there are environments without a window location.

Deraen:
There is no `hostname` option.

Doesn't matter if `host` option can include port, `port` option shouldn't prevent using `host`.

g7s:
@Deraen Sure there is no `hostname` option, I was just proposing it. I pushed another commit that when `port` option is present it overrides the port part of the host or inserts the port if none present.

currentoor:
Also currently React Native users have to set a system property to build their projects, which is undesirable with tools like shadow-cljs which re-use the same JVM process for multiple builds simultaneously.

In this [comment](https://github.com/ptaoussanis/sente/issues/247#issuecomment-555219121), I proposed a solution that doesn't require setting a system property.

ptaoussanis:
Addressing now via https://github.com/ptaoussanis/sente/pull/366 - apologies for the delay on this.

currentoor:
No worries at all @ptaoussanis, thanks for the awesome library!

PR #363, opened 2020-01-25
Allow additional keys in event-msg maps

### LINKED ISSUE #362: Open maps for event-messages
Currently exact keyword matches are enforced for server and client event messages, via the encore function `ks=`:
https://github.com/ptaoussanis/sente/blob/7d8ca800d950026a3171bdd0e438395766b8251a/src/taoensso/sente.cljc#L154

https://github.com/ptaoussanis/sente/blob/7d8ca800d950026a3171bdd0e438395766b8251a/src/taoensso/sente.cljc#L165

Would you consider changing these to `ks>=` to allow sente to be extended in functionality by adding additional items in the map?

My use case is extending sente to allow pub/sub of message. So in addition to a `?reply-fn` there might be a `?pub-fn` that could be executed multiple times. The `?pub-fn` would only be in an event if sent with a `sub-fn` (instead of the usual `send-fn`)

I understand this could be done in user space through slightly different means and/or slight hacks. But it would be nice to be able to call `(add-pub-sub sente-map))` and then be able to use the rest of the sente functions as is.

jjttjj:
Resolves #362

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -151,7 +151,7 @@
 (defn client-event-msg? [x]
   (and
     (map? x)
-    (enc/ks= #{:ch-recv :send-fn :state :event :id :?data} x)
+    (enc/ks>= #{:ch-recv :send-fn :state :event :id :?data} x)
     (let [{:keys [ch-recv send-fn state event]} x]
       (and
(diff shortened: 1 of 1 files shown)


ptaoussanis:
Merged manually, thanks Justin!

PR #365, opened 2020-02-18
Update README.md

mattford63:
CSRF is now mandatory - it took me a whilst to find this out from the official example.   This is quick pass at updating the readme to reflect this.


armincerf:
Actually I don't think CSRF is mandatory, you can pass `:csrf-token-fn nil` in the `make-channel-socket!` options on the server side and it will skip the checks. I don't think that is documented though, I had to dig through the source to find it.

ptaoussanis:
Merging manually, thanks Matt!

PR #366, opened 2020-02-24
Make make-channel-socket-client! respect host option

Rkiouak:
Signed-off-by: Matt Rkiouak <mrkiouak@gmail.com>

PR #370, opened 2020-04-09
Allow user to set a host different from the window location

wilkerlucio:
I tried to use a custom host from an electron app and notice that the host was always going blank. Looking at the code, the `host` param was being shadowed by the internal `host` on the let. This change fixes it so it first tries to use the user-defined `host` instead of reading from the window location.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -1502,8 +1502,8 @@
 
                (let [protocol (or protocol (:protocol win-loc) :http)
                      host     (if port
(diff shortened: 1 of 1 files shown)


wilkerlucio:
@ptaoussanis seems like there is a lot of PR's open, this is very important because not being able to do this setting makes sense unusable in many environments, it seems a simple enough fix on the PR #366 , is that any reason preventing this fix from release? Cheers.

ptaoussanis:
Apologies for the delay on this, have just been swamped. Addressing now - thanks Wilker!

PR #375, opened 2020-07-27
Add Jetty adapter

### LINKED ISSUE #371: jetty adapter
Hi!

First, thanks for this great library!

I wonder why there is no jetty adapter available?
Any reason why it isn't ?

Thanks

wavejumper:
While the official Jetty adapter for Ring does not support websockets, [ring-jetty9-adapter](https://github.com/sunng87/ring-jetty9-adapter) does - which makes this adapter possible.

Fixes #371

I left some notes in the ns docstring, but here's an example  using [ring-defaults](https://github.com/ring-clojure/ring-defaults):

```clojure
(require '[taoensso.sente :as sente])
(require '[taoensso.sente.server-adapters.jetty :as adapters.jetty])
(require '[ring.middleware.defaults :as ring.defaults])
(require '[ring.adapter.jetty9 :as jetty])

(def sente 
  (sente/make-channel-socket! (adapters.jetty/get-sch-adapter) {}))

(def defaults
  (-> ring.defaults/site-defaults
      (assoc-in [:responses :content-types] false)
      (assoc-in [:session :flash] false)))

(jetty/run-jetty handler {:port 3000
                          :websockets {"/chsk" (-> (:ajax-get-or-ws-handshake-fn sente)
                                                   (ring.defaults/wrap-defaults defaults))}
                          :allow-null-path-info true})
```

`:ajax-post-fn` would be configured as a regular route inside of `handler`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@
  * Standard **Ring security model**: auth as you like, HTTPS when available, CSRF support, etc.
  * **Fully documented, with examples**
  * **Small codebase**: ~1.5k lines for the entire client+server implementation
- * **Supported servers**: [http-kit], [Immutant v2+], [nginx-clojure], node.js, [Aleph]
+ * **Supported servers**: [http-kit], [Immutant v2+], [nginx-clojure], node.js, [Aleph], [ring-jetty9-adapter]
 
 ### Capabilities
 
@@ -422,6 +422,7 @@ Copyright &copy; 2014-2016 [Peter Taoussanis].
 [nginx-clojure]: https://github.com/nginx-clojure/nginx-clojure
 [Aleph]: https://github.com/ztellman/aleph
 [example projects]: #example-projects
+[ring-jetty9-adapter]: https://github.com/sunng87/ring-jetty9-adapter
 
 [supported web servers]: https://github.com/ptaoussanis/sente/issues/102
 
--- a/project.clj
+++ b/project.clj
@@ -44,7 +44,7 @@
       [aleph            "0.4.6"]
       [macchiato/core   "0.2.19"]
       [luminus/ring-undertow-adapter "1.1.1"]
-      [wavejumper/ring-jetty9-adapter "0.13.1-patch6"]]}]}
+      [info.sunng/ring-jetty9-adapter "0.13.0"]]}]}
 
   :cljsbuild
   {:test-commands {"node"    ["node" :node-runner "target/main.js"]
--- a/project.clj
+++ b/project.clj
@@ -43,7 +43,8 @@
       [nginx-clojure    "0.5.1"]
       [aleph            "0.4.6"]
       [macchiato/core   "0.2.19"]
-      [luminus/ring-undertow-adapter "1.1.1"]]}]}
+      [luminus/ring-undertow-adapter "1.1.1"]
+      [wavejumper/ring-jetty9-adapter "0.13.1-patch6"]]}]}
 
   :cljsbuild
   {:test-commands {"node"    ["node" :node-runner "target/main.js"]
--- a/src/taoensso/sente/server_adapters/jetty.clj
+++ b/src/taoensso/sente/server_adapters/jetty.clj
@@ -1,0 +1,60 @@
+(ns taoensso.sente.server-adapters.jetty
+  "Sente adapter for ring-jetty9-adapter (https://github.com/sunng87/ring-jetty9-adapter)
+
+  Note: ring-jetty9-adapter requires its own middleware stack for websocket connections.
+
+  Thus, stateful Ring middleware require a little extra configuration:
+
+  * ring-session: pass shared instance of :store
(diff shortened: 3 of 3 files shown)


ptaoussanis:
Awesome, thanks Thomas! Planning to work on Sente this week, will try merge + cut a new release by the weekend.

wavejumper:
@ptaoussanis awesome, thanks! 

And cheers for sente! I love the design, and how simple it was to build an adapter for Jetty. 
We were able to move our product from http-kit to jetty without any fuss 👍

ptaoussanis:
Hi @wavejumper, 

Could you please clarify what you mean by the following in the ns docstring?:

> Note: ring-jetty9-adapter requires its own middleware stack for websocket connections.
>  Thus, stateful Ring middleware require a little extra configuration:  ring-session: pass shared instance of :store.

Do you have a source/reference for this somewhere? I couldn't find anything with a quick Google, and it's not obvious from your code where this might be relevant? Would like to understand better what's actually required of adapter consumers.

Thanks!

wavejumper:
> Hi @wavejumper,
> 
> Could you please clarify what you mean by the following in the ns docstring?:
> 
> > Note: ring-jetty9-adapter requires its own middleware stack for websocket connections.
> > Thus, stateful Ring middleware require a little extra configuration:  ring-session: pass shared instance of :store.
> 
> Do you have a source/reference for this somewhere? I couldn't find anything with a quick Google, and it's not obvious from your code where this might be relevant? Would like to understand better what's actually required of adapter consumers.
> 
> Thanks!

Basically, `ring-jetty9-adapter` defines its websocket routes independently of your regular ring handler. 

In my example above, note that the key `:websockets` has a map of WS paths -> `{"/chsk" ...}`

The caveat is that stateful ring middleware (like `ring-session`) can be instantiated twice.

This is relevant for sente as by default CSRF is enabled, and the default strategy for `ring-anti-forgery` is session based.

A comment made by the author of `ring-jetty9-adapter` demonstrates you can actually reuse the same handler: https://github.com/sunng87/ring-jetty9-adapter/issues/41#issuecomment-630206233 

However, I actually prefer having two independent handlers - as I can configure my middleware stack for websockets to be tailored to API requests, and my regular ring handler tailored to site requests. 

I guess the tl;dr is because there is no unified ring handler, it is easy to configure this Jetty adapter in a way that won't work (eg, you'll get `Bad CSRF token` responses).

More than happy to contribute an example project, or go into detail about usage in the README, just let me know.

ptaoussanis:
Thanks for the extra info!

So if I understand correctly: the configuration that you're talking about here isn't specifically to do with your Jetty adapter - but is just general guidance regarding Jetty 9 and WebSockets?

If this is the case, then I'd prefer not to include too much documentation within Sente or your adapter (since it's a bit out of scope). We can maybe instead just point to a reference (e.g. `ring-jetty9-adapter` README + your attached [link](https://github.com/sunng87/ring-jetty9-adapter/issues/41#issuecomment-630206233))?

wavejumper:
>So if I understand correctly: the configuration that you're talking about here isn't specifically to do with your Jetty adapter - but is just general guidance regarding Jetty 9 and WebSockets?

Correct - the docstring was my attempt to save someone a few hours of pain trying to setup their anti-forgery tokens :)

> If this is the case, then I'd prefer not to include too much documentation within Sente or your adapter (since it's a bit out of scope). We can maybe instead just point to a reference (e.g. ring-jetty9-adapter README + your attached link)?

This sounds like a good idea, I'll update the ns docstring to point to a reference later tonight.

ptaoussanis:
> This sounds like a good idea, I'll update the ns docstring to point to a reference later tonight.

No worries, can do myself quickly - will get PR merged shortly.

Thanks again!

ptaoussanis:
Merged manually. Added a link in the ns docstring to a "full example" Gist at https://gist.github.com/ptaoussanis/5aa2a55f8ca74cdfa77d66e8aa2c9b95.

@wavejumper Please feel free to fork + tweak/extend that example if you like, and I'll mod the link to point to your fork.

Thanks again for the PR, will be great to have Jetty support out-the-box!

wavejumper:
> Merged manually. Added a link in the ns docstring to a "full example" Gist at https://gist.github.com/ptaoussanis/5aa2a55f8ca74cdfa77d66e8aa2c9b95.
> 
> @wavejumper Please feel free to fork + tweak/extend that example if you like, and I'll mod the link to point to your fork.
> 
> Thanks again for the PR, will be great to have Jetty support out-the-box!

Brilliant, thanks!

I'll fork the the gist and extend it to look like how we have Jetty + sente configured for the product I am working on, [Operatr](https://operatr.io)

wavejumper:
@ptaoussanis here's the updated gist that I tested works locally against current master: https://gist.github.com/wavejumper/40c4cbb21d67e4415e20685710b68ea0

PR #385, opened 2021-01-20
Add missing ring-req argument to authorized?-fn function call

viesti:
The documentation on `make-channel-socket-server!` says:

```
    :authorized?-fn    ; ?(fn [ring-req]) -> When non-nil, (authorized?-fn <ring-req>)
                       ; must return truthy, otherwise connection requests will be
                       ; rejected with (unauthorized-fn <ring-req>) response.
                       ;
                       ; May check Authroization HTTP header, etc.
```

But the implementation doesn't actually pass `ring-req` argument to the supplied `authorized?-fn` function. So fixed that by adding the missing argument to the function call :).

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -566,7 +566,7 @@
         (fn [ring-req]
           (if (nil? authorized?-fn)
             false
-            (not (authorized?-fn))))
+            (not (authorized?-fn ring-req))))
 
         ;; nnil if connection attempt should be rejected
         possible-rejection-resp


ptaoussanis:
Merging manually, thanks Kimmo!

viesti:
Thank you! Actually, I ended up doing the check in a middleware that is wrapped around the handlers that sente creates, since I used that middleware elsewhere too. That lead me to think if csrf, or other checks could be delegated to a middleware, maybe then documented also in sente. 

Anyway, this fix is a good thing to keep :)

PR #390, opened 2021-05-18
Add support for websocket-connection options

KaliszAd:
The use-case is e.g. if you want to send larger WS messages. The max-frame-payload is 65536 bytes and max-frame-size is 1048576 bytes by default. It seems, the disassembly/ reassembly of frames isn't supported in some scenarios but there is still need for larger data transfer over WS. Therefore the easiest solution is to just increase the max-frame-payload size. Aleph currently has 7 options to configure more appropriate settings in less common circumstances.

If no map of options is provided, the code works as before and is therefore to the best of my knowledge backwards compatible.

### DIFF
--- a/src/taoensso/sente/server_adapters/aleph.clj
+++ b/src/taoensso/sente/server_adapters/aleph.clj
@@ -24,13 +24,13 @@
   (when-let [s (get-in ring-req [:headers "upgrade"])]
     (= "websocket" (str/lower-case s))))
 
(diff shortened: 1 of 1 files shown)


KaliszAd:
@ptaoussanis would you consider merging the PR or is there anything else I have to do?

ptaoussanis:
Hi Adam, I'm expecting to have time in July to work on open source again- will take a look at this then!

Issue #351, opened 2019-10-24
Websocket and time sensitive CSRF-token - how to update token in state


huima:
Hi,

I am currently wondering a usecase where CSRF-token would be valid only for a limited time. This would become problematic in Sente websocket reconnects, as the same token is used as in the initial handshake.

First thought to my mind would be to solve this by passing new tokens inside the websocket connection as time goes by. As the connection was opened with a valid token, we should trust the user in the other end.

My current test code works as following:

- I take chsk from make-channel-socket-client! and access state available in key :state_

- whenever server pushes updated token to the client, I update the key :csrf-token in state with the new key

I tested also using chsk-reconnect and it seemed that the browser used the updated token as it should.

eneroth:
I'm considering the same question for a different purpose: what happens if the server goes down? When it comes back up, it's going to issue a different CSRF token to the client (unless I somehow persist them, which seems sketchy and hard to do).

A more elegant solution might be for Sente to accept a function as well as a literal CSRF token. If it's a function, it'll evaluate it to get the literal CSRF token. That way, the token can be stored in, for example, the re-frame DB, and updated as needed.

ptaoussanis:
Closing since I'm merging @eneroth's PR in a moment. Thanks Henrik! Apologies for the slow response.

PR #393, opened 2021-11-03
Allow CSRF-token to be a function

eneroth:
- This modifies `make-channel-socket-client!`.
- When the function is evaluated, it's expected to return a CSRF-token.
- CSRF-token strings are still supported.

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -1265,17 +1265,20 @@
 
                     ?socket
                     (try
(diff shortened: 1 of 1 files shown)


eneroth:
This is in reference to https://github.com/ptaoussanis/sente/issues/351.

eneroth:
Note @ptaoussanis, I can update the relevant sections in the readme as well, I just wanted to get your feedback on this before proceeding.

PR #403, opened 2022-05-31
Start ws-kalive loop after connection is established

Snurppa:
Now the :chsk/ws-ping go-loop starts before `(connect-fn)` is called.
If a connection in `(connect-fn)` can't be established right away, it will enter
`(retry-fn)` loop, until WS conn socket is established. Maybe it takes
forever, maybe it takes just 10 seconds for the remote end to come
online.

Meanwhile, we are sending ourselves the pings in ws-kalive-ms intervals.
Basically means log gets polluted with both:
"Chsk send against closed chsk." from the ws-ping send, and
"Chsk is closed: will try reconnect" from the connection `retry-fn`.

This change will start the go-loop only after we have returned from
`(connect-fn)`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ pom.xml*
 *.js
 .env
 .DS_Store
+*.iml
+.idea/
 /lib/
 /classes/
 /target/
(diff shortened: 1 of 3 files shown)


ptaoussanis:
@Snurppa Hi Joni, thanks for this - merging manually now!

PR #409, opened 2022-09-19
Undertow worker pool thread leak caused by ajax fallback requests

kajism:
We were experiencing Undertow worker pool starvation in production. It was caused by ajax requests waiting on never timeouting core.async/<!! channel read. It works better after replacing core.async channel by promise with deref timeout.

### DIFF
--- a/src/taoensso/sente/server_adapters/undertow.clj
+++ b/src/taoensso/sente/server_adapters/undertow.clj
@@ -1,8 +1,8 @@
 (ns taoensso.sente.server-adapters.undertow
-  "Sente server adapter for ring-undertow-adapter."
+  "Sente server adapter for ring-undertow-adapter.
+   Modified to avoid core.async and use promise directly and with read timeout"
   {:author "Nik Peric"}
   (:require
-    [clojure.core.async :as async]
     [ring.adapter.undertow.websocket :as websocket]
     [ring.adapter.undertow.response  :as response]
     [taoensso.sente.interfaces :as i])
@@ -12,6 +12,8 @@
     [io.undertow.websockets
      WebSocketConnectionCallback
      WebSocketProtocolHandshakeHandler]))
+
+(def ajax-response-timeout-ms (* 60 1000))
 
 ;; Websocket
 (extend-type WebSocketChannel
@@ -40,14 +42,18 @@
   (read!  [this])
   (close! [this]))
 
-(deftype SenteUndertowAjaxChannel [ch open?_ on-close]
+(deftype SenteUndertowAjaxChannel [promised-response open?_ on-close]
   ISenteUndertowAjaxChannel
-  (send!  [this msg] (async/put! ch msg (fn [_] (close! this))))
-  (read!  [this] (async/<!! ch))
+  (send!  [this msg] (deliver promised-response msg))
+  (read!  [this]
+    (let [resp (deref promised-response ajax-response-timeout-ms :lp-timed-out)]
+      (close! this)
+      resp))
   (close! [this]
-    (when on-close (on-close ch false nil))
-    (reset! open?_ false)
-    (async/close! ch))
+    (when @open?_
+      (reset! open?_ false)
+      (when on-close (on-close this false nil))
+      true))
 
   i/IServerChan
   (sch-send!  [this websocket? msg] (send! this msg))
@@ -55,9 +61,9 @@
   (sch-close! [this] (close! this)))
 
 (defn- ajax-ch [{:keys [on-open on-close]}]
-  (let [ch      (async/chan 1)
+  (let [promised-response (promise)
         open?_  (atom true)
-        channel (SenteUndertowAjaxChannel. ch open?_ on-close)]
+        channel (SenteUndertowAjaxChannel. promised-response open?_ on-close)]
     (when on-open (on-open channel false))
     channel))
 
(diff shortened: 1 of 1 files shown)


ptaoussanis:
@kajism Hi Karel, thanks for pinging about this- and for the PR. Will add some specific comments shortly.

src/taoensso/sente/server_adapters/undertow.clj
```diff
@@ -1,8 +1,8 @@
 (ns taoensso.sente.server-adapters.undertow
-  "Sente server adapter for ring-undertow-adapter."
+  "Sente server adapter for ring-undertow-adapter.
+   Modified to avoid core.async and use promise directly and with read timeout"
```
ptaoussanis, line 3: Please remove the docstring change here.
The mention of `core.async` seems like an implementation detail, and the read timeout we can better document elsewhere (e.g. in the constructor).

src/taoensso/sente/server_adapters/undertow.clj
```diff
@@ -13,6 +13,8 @@
      WebSocketConnectionCallback
      WebSocketProtocolHandshakeHandler]))
 
+(def ajax-response-timeout-ms (* 60 1000))
```
ptaoussanis, line 16: Would suggest instead adding this as an option to `get-sch-adapter`

src/taoensso/sente/server_adapters/undertow.clj
```diff
@@ -40,24 +42,28 @@
...
   (close! [this]))
 
-(deftype SenteUndertowAjaxChannel [ch open?_ on-close]
+(deftype SenteUndertowAjaxChannel [promised-response open?_ on-close]
   ISenteUndertowAjaxChannel
-  (send!  [this msg] (async/put! ch msg (fn [_] (close! this))))
-  (read!  [this] (async/<!! ch))
+  (send!  [this msg] (deliver promised-response msg))
+  (read!  [this]
+    (let [resp (deref promised-response ajax-response-timeout-ms :lp-timed-out)]
```
ptaoussanis, line 49: Two changes requested here:
1. For compatibility with the old behaviour, please make the timeout optional.
2. Suggest the timeout value be configurable via `get-sch-adapter` option. Maybe a default value of `:undertow/ajax-response-timeout`? (`ajax-response-timeout-val` to better match the `:ajax-response-timeout-ms` option name, and the `undertow` prefix to help clarify that this is an adapter-specific keyword).

src/taoensso/sente/server_adapters/undertow.clj
ptaoussanis: Relatedly: please add a docstring to `get-sch-adapter` to document the 2x new options:

- `:ajax-response-timeout-ms` (and mention its default)
- `:ajax-response-timeout-val` (and mention its default)

`get-sch-adapter` should accept zero args (use default opts), or a single map arg of opts.

Ideally the PR would have 2x commits:
- 1st commit to introduce the new options, but keep the old default behaviour (non-breaking).
- 2nd commit to change the default, which we can mark as potentially breaking.

ptaoussanis (changes requested):
See inline comments, thanks!

kajism:
@ptaoussanis Hello Peter, I have commited the backward compatible suggested changes and will verify it in production.

ptaoussanis:
@kajism Thanks! Content of your PR has been merged on [2022-09-20-pr-409](branch https://github.com/ptaoussanis/sente/commits/2022-09-20-pr-409) and pushed to `[com.taoensso/sente "1.18.0-SNAPSHOT"]` on Clojars.

I don't use Undertow, so it'd be helpful to get your confirmation that this version:
1. Solves the problem
2. Still works correctly otherwise

Once I've got your confirmation, I'll merge with master.

Cheers

kajism:
@ptaoussanis Peter, there seems to be a problem on line 44. I don't see a 3 arity version of clojure.core/deliver ...

kajism:
@ptaoussanis We have moved to Undertow from http-kit after experiencing huge memory leaks in allocated direct memory buffers. After a few days direct memory went up to the max heap size (4GB) and then some short videos on our page stopped playing. The problem is that direct memory is not allocated on the heap, so there was 4GB of heap + 4GB in those buffers.

At first, it was hard to find the the problem, because almost everybody talks only about the heap when speaking about JVM memory. But in the system the number of memory was almost 2 times higher. Also jconsole don't show this memory. It can be found using JMX beans (java.nio:name=direct,type=BufferPool MemoryUsed).

We are sending quite huge ws messages (sometimes more than 1MB), so I was suspecting this may be the reason. After move to Undertow this problem disappeared. In our case, Undertow allocates only up to 90MB in direct memory buffers. But then we experienced the worker thread starvation...

ptaoussanis:
> Peter, there seems to be a problem on line 44. I don't see a 3 arity version of clojure.core/deliver ...

@kajism Apologies, fixed. And updated SNAPSHOT.

kajism:
> > Peter, there seems to be a problem on line 44. I don't see a 3 arity version of clojure.core/deliver ...
> 
> @kajism Apologies, fixed. And updated SNAPSHOT.

Thanks. Locally works fine. Will be deployed in a few days and will confirm then.

kajism:
Works fine. Thanks!

ptaoussanis:
Great, appreciate the confirmation 👍  Cheers

PR #440, opened 2023-10-31
Decrease the log level of a ws-ping event being sent

jwr:
ws-ping events are very frequent and with thousands of connections they tend to overwhelm the logs, bringing little value, even when debugging. They are more suited for the :trace log level, as the other :debug level messages from sente are quite helpful and not overly verbose.


ptaoussanis:
Merging manually now, thanks again Jan! Apologies for the delay.

PR #457, opened 2024-11-18
Make sente compatible with react-native 76

hadils:
Using react-native 76 (Expo 52) and shadow-cljs, js/window hangs in the REPL -- it does not return nil. This causes the UUID code to fail.

Added react-native-target? to check for this. Also turned off event listener code for RN.

Uses a version of encore that has been submitted as a pull request. See https://github.com/taoensso/encore/pull/80

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ pom.xml*
 /checkouts/
 /logs/
 /wiki/.git
+/.clj-kondo/
+/.lsp/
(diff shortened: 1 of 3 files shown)


ptaoussanis:
@hadils Thanks a lot for this Hadil! Apologies for the delayed response.
Will be releasing v1.12 final in the next few hours - and that'll include fixed support for React Native 👍

Issue #325, opened 2018-07-10
Add support for WebSocket requestOptions


Deraen:
In Node (and React-native) the WebSocket constructor can take additional parameter which can set the HTTP request options for the initial request. This can be used to set headers etc.

https://github.com/theturtle32/WebSocket-Node/blob/master/docs/W3CWebSocket.md#constructor
https://github.com/facebook/react-native/blob/29fb2a8e90fa3811f9485d4b89d9dbcfffea93a6/Libraries/WebSocket/WebSocket.js#L97

Inconveniently `W3CWebSocket` and React-Native class take different parameters.

Maybe it would be best to add option to provide a function to create the WebSocket object?

Currently I'm using workaround to replace global WebSocket object:

```clj
(defonce orig-websocket (.-WebSocket goog/global))
(defonce new-websocket
  ;; Note React-native 0.49 changes headers parameters to options
  ;; and headers is property inside options.
  (set! (.-WebSocket goog/global) (fn [url & [protocols headers]]
                                    (let [headers (or headers #js {})]
                                      (gobj/set headers "X-Access-Key" config/api-access-key)
                                      (new orig-websocket url protocols headers)))))
```

ptaoussanis:
@Deraen Hi Juho, apologies for the long delay replying!
An option for this will be included in forthcoming v1.18 release.

Cheers!

Issue #418, opened 2023-02-23
Pass CSRF token using `Sec-WebSocket-Protocol` header


ptaoussanis:
For more context: https://github.com/ptaoussanis/sente/issues/137#issuecomment-542281213

ptaoussanis:
Just a heads-up that v1.22 [includes](https://github.com/taoensso/sente/releases/tag/v1.22.0-RC1) more flexible connection rejection logic - may be helpful here 👍

PR #464, opened 2025-04-02
Fix option to use custom WebSocket constructor

iarenaza:
Commit f560294 introduced the option to use a custom Websocket constructor. But even if the caller of `make-channel-socket-client!` passed in the optional `:ws-constructor` key, the value of that key was never used. The code from that commit always used the `default-client-ws-constructor` function.

Which probably means nobody has ever used that option until now :-)

[Re: #325]

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -1316,7 +1316,8 @@
                   (enc/oget @?node-npm-websocket_ "w3cwebsocket"))]
 
        (delay
-         (let [socket (WebSocket. uri-str)]
+         (let [protocols (:sec-websocket-protocol headers)
+               socket (WebSocket. uri-str protocols)]
            (doto socket
              (aset "onerror"   on-error)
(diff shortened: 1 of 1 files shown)


ptaoussanis:
@iarenaza Thanks for the clear issue and fix Iñaki! 🙏 Do you need a release with this fix?

iarenaza:
I have another patch for issue #418 (that I discovered after I tried to use the '':ws-constructor` option for that same purpose, and failed for obvious reasons :smile: ). But I need to polish the commit message a bit. So maybe it's a better idea to do a release once I send the pull request for #418 ?

iarenaza:
@ptaoussanis It seems Github Pull Request feature doesn't let me create a separate pull request for the  #418 issue until you merge this one. Otherwise, it merges the two commits in this same pull request (which is less than ideal).

So I'll wait until you merge this one, to open the other PR for #418.

ptaoussanis:
I've manually merged your fix 👍 Please note that you'll need to rebase on master for #418.

PR #465, opened 2025-04-02
Pass CSRF token using Sec-WebSocket-Protocol header

iarenaza:
The proposed change tries to respect other values in the Sec-WebSocket-Protocol header, that the caller may have set in the `headers` option when calling `make-channel-socket-client!`

[Re: #418]

### DIFF
--- a/src/taoensso/sente.cljc
+++ b/src/taoensso/sente.cljc
@@ -277,6 +277,20 @@
   (allow-origin? #{"http://site.com"} {:headers {"referer" "http://attacker.com/"}})
   (allow-origin? #{"http://site.com"} {:headers {"referer" "http://site.com.attacker.com/"}}))
 
+(def ^:private sente-csrf-token-prefix "sente-csrf-token-")
+
+(defn- sente-csrf-token-pred [s]
+  (when (str/starts-with? s sente-csrf-token-prefix)
+    (subs s (count sente-csrf-token-prefix))))
+
+(defn- ws-csrf-token [ring-req]
+  (let [ws? (= "websocket" (get-in ring-req [:headers "upgrade"]))
+        sec-websocket-protocol (get-in ring-req [:headers "sec-websocket-protocol"])
+        protocol-vals (when (and ws? (string? sec-websocket-protocol))
+                        (-> sec-websocket-protocol
+                            (str/split #", *")))]
+    (some sente-csrf-token-pred protocol-vals)))
+
 (defn valid-csrf-token?
   "Returns true iff given Ring request has valid CSRF token as per `csrf-token-fn`
   (see [[make-channel-socket-server!]] `:csrf-token-fn` option).
@@ -297,7 +311,8 @@
             (or
               (get-in ring-req [:params    :csrf-token])
               (get-in ring-req [:headers "x-csrf-token"])
-              (get-in ring-req [:headers "x-xsrf-token"]))]
+              (get-in ring-req [:headers "x-xsrf-token"])
+              (ws-csrf-token ring-req))]
         (boolean
           (enc/const-str=
             ref-token
@@ -736,6 +751,16 @@
                           :?reply-fn ?reply-fn
                           :uid       uid}))))
 
+                  pre-ws-handshake
+                  (fn [server-ch ring-req]
+                    (if-let [csrf-token (ws-csrf-token ring-req)]
+                      {:do-handshake true
+                       :handshake-extra-headers {"Sec-WebSocket-Protocol"
+                                                 (str sente-csrf-token-prefix csrf-token)}}
+                      {:do-handshake false
+                       :no-handshake-ring-resp {:status 400
+                                                :body "Invalid Sec-WebSocket-Protocol header."}}))
+
                   send-handshake!
                   (fn [server-ch websocket?]
                     (trove/log!
@@ -969,7 +994,8 @@
                  :on-open             on-open
                  :on-msg              on-msg
                  :on-close            on-close
-                 :on-error            on-error}))))))}))
+                 :on-error            on-error
+                 :pre-ws-handshake    pre-ws-handshake}))))))}))
 
 (def ^:dynamic *simulated-bad-conn-rate*
   "Debugging tool. Proportion ∈ℝ[0,1] of connection activities to sabotage."
@@ -1251,7 +1277,8 @@
                   (enc/oget goog/global           "MozWebSocket")
                   (enc/oget @?node-npm-websocket_ "w3cwebsocket"))]
        (delay
-         (let [socket (WebSocket. uri-str)]
+         (let [protocols (or (:sec-websocket-protocol headers) [])
+               socket (WebSocket. uri-str protocols)]
            (doto socket
              (aset "binaryType" binary-type)
              (aset "onerror"    on-error)
@@ -1483,13 +1510,18 @@
                             {:on-error   on-error
                              :on-message on-message
                              :on-close   on-close
-                             :headers    headers
+                             :headers    (update headers :sec-websocket-protocol
+                                           (fn [x]
+                                             (let [csrf-token (str sente-csrf-token-prefix
+                                                                   (get-client-csrf-token-str :dynamic (:csrf-token @state_)))]
+                                               (cond
+                                                 (string? x) [x csrf-token]
+                                                 (coll? x) (conj x csrf-token)
+                                                 :else csrf-token))))
                              :uri-str
                              (enc/merge-url-with-query-string url
                                (merge params ; 1st (don't clobber impl.):
-                                 {:client-id  client-id
-                                  :csrf-token (get-client-csrf-token-str :dynamic
-                                                (:csrf-token @state_))}))}))
+                                 {:client-id  client-id}))}))
 
                         (catch :any t
                           (trove/log! {:level :error, :id :sente.client/ws-constructor-error, :error t}))))]
--- a/src/taoensso/sente/server_adapters/http_kit.clj
+++ b/src/taoensso/sente/server_adapters/http_kit.clj
@@ -18,14 +18,15 @@
 (deftype HttpKitServerChanAdapter []
   i/IServerChanAdapter
   (ring-req->server-ch-resp [sch-adapter ring-req callbacks-map]
-    (let [{:keys [on-open on-close on-msg _on-error]} callbacks-map
+    (let [{:keys [on-open on-close on-msg _on-error pre-ws-handshake]} callbacks-map
           ws? (:websocket? ring-req)]
 
       ;; Note `as-channel` requires http-kit >= v2.4.0
       ;; Returns {:body <http-kit-implementation-channel> ...}:
       (hk/as-channel ring-req
-        {:on-close   (when on-close (fn [sch status-kw] (on-close sch ws? status-kw)))
-         :on-receive (when on-msg   (fn [sch       msg] (on-msg   sch ws? msg)))
-         :on-open    (when on-open  (fn [sch          ] (on-open  sch ws?)))}))))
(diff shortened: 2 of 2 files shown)


iarenaza:
I opted for prefixing the actual CSRF token value with `sente-csrf-token-`, to make it easier to find it and extract it on the server side. One of the reasons for that is because protocol names in the Sec-WebSockel-Protocol header cannot use a bunch of non-alphanumeric characters (like spaces, colons, dashes, etc). And we want the value to be easily recognizable, in case the client wants to send other protocol names in that same header (for some custom extra behavior).

ptaoussanis:
@iarenaza Thank you Iñaki, this looks good! Will try take a closer look + merge by end of tomorrow 👍

iarenaza:
Whoops. I had forgotten the other side of the connection: extracting the CRSF token from the Sec-WebSocket-Protocol header on the server side.

ptaoussanis:
@iarenaza Hi Iñaki, sorry for the delay on this.

Before merging, I'd like to step back a moment and re-evaluate what we're trying to do here. The current PR seems to be based on [this](https://github.com/taoensso/sente/issues/137#issuecomment-542281213) suggestion made earlier by @kaosko, right?

But it's actually not obvious to me what the motivation is/was behind that suggestion.

@kaosko Sorry for digging this up after so long, but could you maybe explain why you've previously suggested that the Sente CSRF token be sent in the `Sec-WebSocket-Protocol` header? (Context [here](https://github.com/taoensso/sente/issues/137#issuecomment-542281213)). It's not obvious to me what the benefit is, or that the [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol) is intended as a place to convey a CSRF token.

Even though the Protocol header seems to be an application-level header, it seems we'd be abusing the semantic intention of the header no? I imagine that stands a non-zero chance of causing issues in the future, and seems like it anyway could be surprising. What's the anticipated upside?

kaosko:
It's a JavaScript spec/API thing. It doesn't allow you to set any other headers for the WebSocket request object than "Sec-WebSocket-Protocol" so [if you want to pass auth/CSRF tokens in the headers, that's the way to do it](https://stackoverflow.com/questions/22383089/is-it-possible-to-use-bearer-authentication-for-websocket-upgrade-requests/35108078#35108078). I wouldn't call it abusing the semantic intention, perhaps merely stretching it, again as an _application_ level protocol. In the grand scheme of things, I guess not that different from manufacturer specific Wi-Fi Information Elements or DHCP Vendor Options.

Stepping back, [the anticipated upside according to OWASP (the most authoritative I could find) is that "A CSRF token must not be leaked in the server logs or in the URL"](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern).

Almost anything has a non-zero chance of causing issues in the future, but the server is certainly free to ignore any values in Sec-WebSocket-Protocol that it doesn't understand. Plenty of web pages, frameworks still sending various "X-*" headers while the naming practice was deprecated in 2012 and the servers may or may not act on those (and vice versa). That said, I'm not strongly arguing for adding this but don't see how it could hurt much either. Finally, I have used and still use the Sec-WebSocket-Protocol header for sending authentication tokens with the websocket request.

iarenaza:
@ptaoussanis Just to confirm that `1.21.0-alpha1` works as expected in our case :tada: . We no longer need to use the `:ws-custom` option on the client side :pray: 

Side comment: we still need to use some of the available "callback functions" (like `csrf-token-fn`) on the backend side (to deal with server side CSRF token retrieval and validation). But that is totally expected, because of the type of value that we are using as the CSRF token (as I already mentioned, we are using OpenID Connect ID tokens as the CSRF token)

So, thanks a lot @ptaoussanis !

ptaoussanis:
Great, thanks for the confirmation @iarenaza!

kaosko:
What about Aleph, Undertow etc? It seems that http-kit just returns [whatever you stick in :ring.websocket/protocol key of the ring response](https://github.com/http-kit/http-kit/blob/master/src/org/httpkit/server.clj), no?

ptaoussanis:
@iarenaza Hi Iñaki, thanks for the detailed+clear update 🙏

> P.S. We are so sorry that we didn't test it with Chrome before submitting this PR and wasted your valuable time 😞

No problem at all, I appreciate the time + effort you've put into this! I'm sorry that you've been running into issues with Sente and/or http-kit!

Re: how to proceed- I've unfortunately got limited knowledge + bandwidth on this topic since I'm currently quite deeply buried in a few other things.

I am planning a big batch of Sente work for June, but hopefully we can make progress here in the meantime!

Can we first back up a bit so that I can confirm my understanding?

@iarenaza Your motivation for #465 is described [here](https://github.com/taoensso/sente/pull/465#issuecomment-2782694714), right?

So:

- You already have a value that'd work as a CSRF token.
- But you don't want the client to send it to the server over a query param since you're concerned about the token leaking / getting logged.

So the goal is to get the CSRF token from the client to the server in a way that:

1. Doesn't leak the token.
2. Works reliably across different browsers and HTTP servers.
3. Minimizes the size/complexity of changes needed in Sente.

Is that right?

If we were to enumerate all the options we have for getting this done- what would that include?

1. Send the token in a `Sec-WebSocket-Protocol` header
2. Send the token in another header? Is that possible?
3. Send the token in the first WebSocket message? Is that possible?
4. Any other ideas?

I haven't had the opportunity to properly investigate this myself, so any input here would be very welcome!

Put briefly, what's currently the main impediment to (1)?

What do other/similar libraries tend to do?

iarenaza:
Hi @ptaoussanis ,

I have finally extracted my original changes from my local test repos, cleaned them up and removed all the `prn`s, etc. And I have applied those changes on top of the sente and http-kit main/master branches (as of today, 2025.11.26). I have kept the original sente and http-kit versions in their respective `project.clj` files, to be able to test them more easily using the Leiningen "checkouts" feature.

Then I have created a separate git repo, using the `example-project` code from the sente project, where I have cloned my forked repositories in the `checkouts` directory.

This is what I have done:

1. Clone the https://github.com/iarenaza/sente-example-project.git repository locally.

2. Create a `checkouts` directory inside the cloned repository.

3. Inside the `checkouts` directory, clone these two other repositories, that contain my patches:

   - https://github.com/iarenaza/sente.git
   - https://github.com/iarenaza/http-kit.git

4. After the previous steps, we should have the following directory structure:

   ```
   .
   ├── checkouts
   │   ├── http-kit
   │   └── sente
   ├── resources
   │   └── public
   └── src
       └── example
   ```

5. Now we can run `lein start` or `lein start-dev`, as in the original `example-project` from the sente project, from the top of the `sente-example-project` cloned repository.

6. I tested it with both Firefox 144.0 and Chromium "142.0.7444.175 (Official Build) built on Debian GNU/Linux 13 (trixie) (64-bit)", both running on Debian Trixie, and it worked as expected.

Do the changes look like something that you could consider including?

ptaoussanis:
@iarenaza Hi Iñaki, sorry for the delay replying- this had caught me right in the middle of a big move. Am settled again 👍

I see you updated the branch, thanks for that.

So just to confirm my understanding:

- The idea is to introduce a new pre-ws-handshake option to http-kit and to server adapters, is that right?

- Does this have any downside for users of older versions of http-kit?
- Does this have any downside for users of other (non http-kit) HTTP servers?
- Any idea if any other HTTP servers actually have a similar capability?
- Have you been running this in production successfully?
- What's your level of satisfaction with proposal as-is? Any significant open questions or trade-offs that need to be discussed?

Cheers! :-)

iarenaza:
> @iarenaza Hi Iñaki, sorry for the delay replying- this had caught me right in the middle of a big move. Am settled again 👍
> 
> I see you updated the branch, thanks for that.
> 
> So just to confirm my understanding:
> 
> * The idea is to introduce a new pre-ws-handshake option to http-kit and to server adapters, is that right?

To be honest, I haven't explored the other server adapters, so this may or may not be needed.

The current patch addresses a (IMHO) limitation of http-kit Websocket connections management (when using the `as-channel` mechanism and associated protocol[1], I haven't had a look at the code path that uses the ring websocket protocols).

Mainly that there is no way to reject the websocket connections (much less with a custom response). Or better said, to conditionally accept or reject them. But also that there is no way to add any details to the headers or the body of the responses (irrespective if it is to accept or reject the connection).

[1] Which is what Sente uses with the http-kit adapter.

> * Does this have any downside for users of older versions of http-kit?

Well, it certainly would have downsides. Older versions of http-kit would not know about the pre-ws-handshake option. And thus the (proposed) sente code that uses the websocket specific headers to send the CSRF token would not work.

It would probably not fail in the sense of throwing exceptions (though I haven't analyzed this particular scenario yet, to be honest). But the end result would be that the websocket connection attempts would be rejected[2]. Because the relevant websocket response headers wouldn't be there in that case.

[2] At least in Chrome-based browsers. Firefox, as of today would probably work. But this could change at any moment, because, as we discusses in this same issue some months ago, that part of the spec seems to be somewhat open to interpretation.

> * Does this have any downside for users of other (non http-kit) HTTP servers?

To be honest, I don't know. I haven't spent time on that, because I wanted to see if you were open to this approach at all, before committing to a non-trivial amount of research :-)

> * Any idea if any other HTTP servers actually have a similar capability?

I know "generalist" HTTP servers like nginx and Apache have support for this, because they basically expose lots of low level details.

But speaking of JVM based HTTP servers, like Jetty, etc, I don't know. But ring.websocket.protocols doesn't seem to offer anything like the pre-ws-handshake feature. So I guess the underlying Jetty adapter doesn't either? And quite probably other JVM based HTTP servers also don't have anything like this.

> * Have you been running this in production successfully?

We haven't deployed this to production. We are still using the stock http-kit + sente versions, with the CSRF token sent in the URL as a query param.

I am not "attached" to this approach, in the sense that any other approach that achieves the same final result (not sending the CSRF token in the connection URL) is equally fine with me.

I simply thought that rejecting the websocket connections before they are fully established (the upgrade handshake as been completed) made more sense then finishing the handshake only to close the connection after the first message was received.

Which, as a side-effect, also would also reduce the impact of http-kit issues http-kit/http-kit#473 and http-kit/http-kit#566.

But I can definitely see how all of this makes things harder for users of other HTTP servers, or older versions of http-kit.

> * What's your level of satisfaction with proposal as-is? Any significant open questions or trade-offs that need to be discussed?

As I said in my previous comment in this issue, the proposal's goal was to get this discussion going based on something concrete. And see if you had interest in any of this, or thought it was to convoluted, or too problematic for other use cases (like the one you brought up about compatibility with older http-kit versions)

Other than that, code-wise, I'm reasonably satisfied the implementation. But, as I said above, I'm open to other approaches that you may think are better for a wider audience.

Best regards,

ptaoussanis:
@iarenaza Hi Iñaki,

> I haven't spent time on that, because I wanted to see if you were open to this approach at all, before committing to a non-trivial amount of research :-)

Gotcha, thanks for clarifying- and for all your work on this 👍

So in its current form, it's difficult to judge. I'm not opposed to the idea in principle, but unfortunately would need the non-trivial details in place to be able to properly evaluate.

The short / high-level version is this:

- I'm open to sensible non-breaking additions/improvements to http-kit and Sente.
- In both cases also _potentially_ open to breaking changes iff there's a strong  security/correctness need, and a very strong argument for why non-breaking alternatives wouldn't be sufficient.

So the key questions I'd want addressed before I'd be able to say if the path is viable or not:

* Is there a way we could do this without breaking anyone?
* Is there a way we could do this while also offering support to other servers?

If it's very costly to determine the above definitively, then it'd at least be helpful as a starting point to estimate.

Hope that's some help at least!

PR #466, opened 2025-04-11
fix for jetty 12

aiba:


aiba:
@ptaoussanis do you think you could take a look at this very simple fix, and perhaps cut a new sente release?

It's hard to deploy this patch ourselves because it looks like the sente project doesn't have a manifest that enables depending on it from deps.edn using {:git/url ...} coordinates.

ptaoussanis:
@aiba Hi Aaron, of course! Apologies, didn't realise that you were waiting for a release on this. Just finishing something up, then will cut a new release (w/in the next hour) 👍

Thanks for the PR! 🙏

PR #468, opened 2025-06-07
Update wiki that `wrap-session` is broken for reitit users

kawpuh:


ptaoussanis:
@kawpuh Hi Ethan, thanks for pinging about this - and for the PR!

I'm not really sure that this warning belongs in the Sente docs TBH. The Sente example doesn't use or otherwise mention retit, and this seems to be a general issue on retit's end - so would maybe better be mentioned in the retit's docs rather than in the docs of individual libraries that people _might_ use with retit?

PR #474, opened 2025-10-01
Add UUID support to msgpack

jrgvf:
If we try to pack a UUID, we will receive an error (StackOverflowError in Clojure). The idea is to add the necessary extension to support UUID when using msgpack as the Packer.


ptaoussanis:
Thanks for this! Am expecting to cut a new release in a couple weeks, will get this included then 👍

